Issue #6, opened 2015-02-11
Dragging can't actually be disabled.


stevemckenzie:
You have a props.draggable but it's never actually used anywhere.

akiran:
Fixed this in commit 837466b02a744ae4b9e723f440379b861c379c1f

swipe = false, will disable swipe or dragging feature on both mobile and desktop.
swipe = true & draggable = false will disable dragging on desktop devices and swipe works on mobile.

benkauffman:
what if I wanted to disable dragging on mobile but enable it for desktop?

antkiewiczk:
@benkauffman you can still use responsive settings, same as with e.g slidesToShow

PR #8, opened 2015-02-11
Only allow dragging when draggable.

stevemckenzie:


stevemckenzie:
https://github.com/akiran/react-slick/issues/6

akiran:
@stevemckenzie  
according to [slick](http://kenwheeler.github.io/slick/)
draggable  is to enable mouse dragging (for desktop devices)
swipe is to enable swiping (for mobile devices)

With draggable = false and swipe = true , draggable feature should be off on desktop and should work on mobile devices.

You fix turns of draggable on mobile devices  also.
Can you make changes to consider swipe option also ?

akiran:
@stevemckenzie  Thanks for PR.
This on PR only fixes the issue partially

Fix for #6 is applied in commit 837466b

Issue #7, opened 2015-02-11
Include only necessary methods in lodash


akiran:
As of now, whole lodash module is imported.
Import only necessary methods to reduce file size.

PR #9, opened 2015-02-13
[issue#7] Includes only necessary methods of lodash

edygar:
Also, I put "use strict" accordingly to the `.jshintrc`.

This pull requests solves https://github.com/akiran/react-slick/issues/7


akiran:
@edygar Thanks for pull requested

PR #26, opened 2015-03-04
Enable viewport scrolling on touch devices

grrowl:
Only when a significant horizontal movement hasn't been made.

### DIFF
--- a/lib/mixins/event-handlers.js
+++ b/lib/mixins/event-handlers.js
@@ -53,7 +53,6 @@ var EventHandlers = {
         curY: posY
       }
     });   
-    e.preventDefault();
   },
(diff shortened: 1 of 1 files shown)


quarklemotion:
These changes also fix an issue I was having where tap/click events were not firing for links within the carousel divs.  What appears to fix this is moving the preventDefault in swipeEnd() until after `if (!touchObject.swipeLength) { return; }`, as swipeLength will be undefined for a tap with no dragging movement.

akiran:
@grrowl  This seems to fix #38 
I couldn't merge this because of merge conflicts.

But I copied the changes and it worked.
Thanks for PR

PR #28, opened 2015-03-12
Fix autoplay behaviour

grrowl:
Now waits *n* ms since last action. Any slide transition resets the timer.


akiran:
@grrowl Thanks a lot for this pull request.

akiran:
Released 0.3.8 with this fix.

PR #30, opened 2015-03-19
Prevent touches from scrolling if touchMove=false

yuzeh:
This is implemented in pretty much the same way that `draggable = false` is.

I tested this on an app I'm writing that depends on this code. Let me know if there is a better way to test.


akiran:
To disable touch swipe, you can set swipe props as false.
swipe = false feature is already working with react-slick

PR #32, opened 2015-03-23
Adding the ability to specify custom next and prev arrow elements

ryanbagwell:
This allows you to pass in custom next and prev elements as props.

### DIFF
--- a/lib/default-props.js
+++ b/lib/default-props.js
@@ -32,8 +32,8 @@ var defaultProps = {
     vertical: false,
     // waitForAnimate: true,
     afterChange: null,
-    nextButton: null,
-    prevButton: null
+    nextArrow: null,
+    prevArrow: null
 };
 
 module.exports = defaultProps;
--- a/lib/default-props.js
+++ b/lib/default-props.js
@@ -31,7 +31,9 @@ var defaultProps = {
     variableWidth: false,
     vertical: false,
     // waitForAnimate: true,
-    afterChange: null
+    afterChange: null,
+    nextButton: null,
+    prevButton: null
 };
 
 module.exports = defaultProps;
--- a/lib/inner-slider.jsx
+++ b/lib/inner-slider.jsx
@@ -100,6 +100,7 @@ var Slider = React.createClass({
     );
   },
   renderArrows: function () {
+
     if (this.props.arrows === true) {
       var prevClasses = { 'slick-prev': true};
       var nextClasses = { 'slick-next': true};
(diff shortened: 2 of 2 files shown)


akiran:
I will review this pull request soon

ryanbagwell:
Thanks. I new to react, so I'm not sure if this is the best way to accomplish this. It requires something like this:

````
var prevArrow = React.createClass({

  render: function() {
    return <button {...this.props}>Previous</button>
  }

})

<Carousel prevArrow=prevArrow>
  <div>One</div>
  <div>Two</div>
  <div>Three</div>
</Carousel>
````

Note that ````this.props```` has to be passed to the element. At the least, I probably should have added an update to the documentation.

**Comment by ryanbagwell** [comment]
@akiran just wanted to see what you thought about this PR.

**Comment by akiran** [comment]
@ryanbagwell, Sorry. I couldn't get to this because of other tasks.
I will definitely review this PR and merge it this weekend.

**Comment by ryanbagwell** [comment]
awesome, thank you.

File: lib/default-props.js [hunk 1]
```diff
@@ -31,7 +31,9 @@ var defaultProps = {
     variableWidth: false,
     vertical: false,
     // waitForAnimate: true,
-    afterChange: null
+    afterChange: null,
+    nextButton: null,
```
akiran: @ryanbagwell  Can you change these names to "nextArrow" and "prevArrow"  to match the terminology of  slick carousel

lib/inner-slider.jsx
```diff
@@ -127,8 +128,36 @@ var Slider = React.createClass({
...
+        key: '0',
+        ref: 'previous',
+        'data-role': "none",
+        className: classnames(prevClasses),
+        style: {display: 'block'},
+        onClick: prevHandler
+      }
+
+      if (this.props.prevArrow) {
+        var prevArrow = <this.props.prevArrow {...prevArrowProps} />;
```
akiran: Here you used the prevArrow and nextArrow terminology correctly.
I think you made a mistake in default-props.js file.
Can you correct it.

ryanbagwell:
@akiran sorry about that. I pushed a fix. Thanks for letting me know.

akiran:
@ryanbagwell  Thanks for the pull request.

chemoish:
Is there an example implementation of this?

ryanbagwell:
I should have included one in the readme. You basically pass in react components as props with names prevArrow and nextArrow.

chemoish:
I don't believe there is any documentation and I have yet to reverse engineer the source https://github.com/akiran/react-slick/commit/cfb1c3212c50e34a7e4df1b226447e86db9b488c.

I have tried

```js
// inline
<Slick nextArrow={<button></button>}></Slick>

// settings
var settings = {
  nextArrow={<button></button>}
};

<Slick {...settings}></Slick>
```

Seems that https://github.com/akiran/react-slick/blob/44ae518cd08fc4fd7d5523738591832a5e4d5711/src/arrows.jsx#L31 only accepts a string to modify the tag name (which doesn't seem that useful).

---

Anyways, I am looking into https://github.com/akiran/react-slick/blob/44ae518cd08fc4fd7d5523738591832a5e4d5711/src/inner-slider.jsx#L78, but figured I would ask.

ryanbagwell:
Your example should work. Not sure if you need brackets around your button component.

ryanbagwell:
@chemoish see my comment on March 24 and note the part about passing this.props to your button component.

chemoish:
@ryanbagwell, I tried your code, but it doesn't seem to satisfy the use case (Correct me if I am wrong—Thanks for the response btw).

```js
let NextArrow = React.createClass({
    render() {
        return (
            <button className="slick-arrow" {...this.props}>Next</button> // need a slick-arrow class
        );
    }
});
```

The current implementation allows you to override the `tag` as well as the `innerHTML`. It also allows you to add default props (Both inline or configured).

However, it doesn't allow you to extend properties that are hard coded in https://github.com/akiran/react-slick/blob/44ae518cd08fc4fd7d5523738591832a5e4d5711/src/arrows.jsx#L69.

---

Solution 1: Copy pasta https://github.com/akiran/react-slick/blob/44ae518cd08fc4fd7d5523738591832a5e4d5711/src/arrows.jsx#L42 and pass that in as a prop (Bad).

Solution 2: ???

PR #33, opened 2015-03-23
Change this.props.initialSlide to props.initialSlide in initialize

realchaseadams:
props is passed into initialize and initialize is not being bound to the places it's called so using props instead of this.props gives the correct initialslide for the current props being passed in.

This causes a problem of if you're passing in initialSlide, and whenever a state changes out of the component, the component will be one slide behind because it was using the previously set this.props.initialSlide instead of the current props.initialSlide passed in.


akiran:
@realchaseadams  Thanks for pull request.

Issue #25, opened 2015-03-04
iOS: Can only swipe once, doesn't transition (state.animating locks to true)


grrowl:
The problem is from line 203 of `lib/mixins/helpers.js`

```js
    this.setState({
      animating: true,
      currentSlide: currentSlide,
      currentLeft: currentLeft,
      trackStyle: this.getAnimateCSS(targetLeft)
    }, function () {
      ReactTransitionEvents.addEndEventListener(this.refs.track.getDOMNode(), function() {
        this.setState({
          animating: false,
          trackStyle: this.getCSS(currentLeft),
          swipeLeft: null
        });
      }.bind(this));
    });
```

For whatever reason iOS doesn't get the css `transition` class (either not in time or doesn't support?), therefore the `ReactTransitionEvents.addEndEventListener` never fires and `animating` is stuck to `true`, which blocks subsequent actions.

akiran:
Thanks for posting the issue.

This week I am little busy. I will fix all the pending issues of react-slick next week.

grrowl:
No worries, I'm applying little workarounds to my local build (in this case, commenting out the line `animating: true` :sweat_smile:). 

Also, here's a repro video: https://www.dropbox.com/s/uzesxuacxur7qnc/react-slick-ios.mov?dl=0

grrowl:
Confirmed the `transition` class doesn't get applied in time. (adding a permanent `transition: transform;` works as expected.

also, this issue causes `afterChange()` not to fire on iOS at all.

akiran:
Thanks, I will review this.

kriswallsmith:
I'm seeing this issue also and may not be able to use this module because of it… shame because it's great work otherwise!

grrowl:
If you're okay with transitions not working on iPhone (but otherwise functional), you can add [my fork](https://github.com/grrowl/react-slick/tree/bugfixes-and-bowerless) to your package.json:
`  "react-slick": "grrowl/react-slick#bugfixes-and-bowerless",`

akiran:
If animating state is stuck to true, then for some reason callback attached to transition event is not called.

grrowl:
That's the fix, in the fork we skip setting `state.animating` to true, so it doesn't matter is the callback is fired or not. I haven't noticed any drawbacks (apart from the effect being kinda kludgy on iOS)

[...]

PR #35, opened 2015-03-25
Apply a -webkit-transition for the -webkit-transform.

grrowl:
Otherwise iOS 8 Safari is transitioning `transform`, but applying `-webkit-transform`
(and ignoring `transform`).


grrowl:
Fixes #25

grrowl:
Also, tested in Chrome stable, iOS 8.1 Safari and Desktop Safari 8.0.3

akiran:
@grrowl.  Huge thanks.
Just tested this on iphone.  Working well.

PR #36, opened 2015-03-27
Added customPaging option mimicking behavior found in slick itself.

rockymadden:
There are slight differences from the slick implementation:

* Rather than pass in `this` as the first argument to the `customPaging` function and make the developer handle the in-and-outs of building the onClick event, I simply provide a function that can be passed directly to onClick without them having to worry (there is also an issue where you would have to pass `dotOptions` and they would have to pass it back, if this approach was not taken). 
  * `<button onClick={onClick}>` vs `<button onClick={this.changeSlide.bind(this, dotOptions)}>`
* In addition to the index, the `customPaging` function is also passed the react element that corresponds to each dot. This would allow you, for instance, to create descriptive buttons rather than dots based upon information held in the react element and its children (e.g. image title as button text).


akiran:
This feature is added in 0.14.3

PR #55, opened 2015-05-25
Better formatting + ES6

vramana:


akiran:
@vramana I think we should not class syntax of react to support react 0.12 users.
We can use it in future.

Can you revert the changes to React.createClass

vramana:
Reverted.

vramana:
@akiran This line https://github.com/akiran/react-slick/blob/master/lib/initial-state.js#L32 
` // $prevArrow: null,` is under removed.
 So, is this line https://github.com/akiran/react-slick/blob/master/lib/inner-slider.jsx#L141 relevant ??
`var prevArrow = <this.props.prevArrow {...prevArrowProps} />;`

akiran:
Yes  https://github.com/akiran/react-slick/blob/master/lib/inner-slider.jsx#L141  is relevant.   It default value is at https://github.com/akiran/react-slick/blob/master/lib/default-props.js#L36

PR #57, opened 2015-05-25
[WIP] Fix warnings in demos caused by render function being not pure

vramana:
- [x] Center Mode (3/4)
- [x] Adaptive Height (1/4)

### DIFF
--- a/lib/mixins/helpers.js
+++ b/lib/mixins/helpers.js
@@ -152,15 +152,14 @@ var helpers = {
       'slick-cloned': slickCloned
     });
   },
-  getListStyle: function () {
-    var style = {};
+  adaptHeight: function () {
     if (this.props.adaptiveHeight) {
       var selector = '[data-index="' + this.state.currentSlide +'"]';
       if (this.refs.list) {
-        style.height = this.refs.list.getDOMNode().querySelector(selector).offsetHeight;
+        var slickList = this.refs.list.getDOMNode();
+        slickList.style.height = slickList.querySelector(selector).offsetHeight + 'px';
       }
     }
-    return style;
   },
   slideHandler: function (index, sync, dontAnimate) {
     // Functionality of animateSlide and postSlide is merged into this function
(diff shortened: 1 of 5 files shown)


lib/mixins/helpers.js
```diff
@@ -119,31 +119,18 @@ var helpers = {
...
-    var allSlides;
-    var centerIndex;
-    var realRange = false;
-
+    var currentSlide = this.state.currentSlide;
+    var centerOffset;
 
     slickCloned = (index < 0) || (index >= this.state.slideCount);
     if (this.props.centerMode) {
-      if (this.refs.track) {
```
vramana: @akiran I simplified the implementation here and removed all the unnecesary variables. Have a look at it.

vramana: Visual representation
atom_2-2

lib/mixins/helpers.js
```diff
@@ -152,15 +139,14 @@ var helpers = {
...
   },
-  getListStyle: function () {
-    var style = {};
+  adaptHeight: function () {
     if (this.props.adaptiveHeight) {
       var selector = '[data-index="' + this.state.currentSlide +'"]';
       if (this.refs.list) {
-        style.height = this.refs.list.getDOMNode().querySelector(selector).offsetHeight;
+        var slickList = this.refs.list.getDOMNode();
+        slickList.style.height = slickList.querySelector(selector).offsetHeight + 'px';
```
akiran: @vramana Did you make this change to avoid react 0.13 warning that getDOMNode() is called from render method

vramana: Yeah. The warning said getDOMNode operation should be done inside `componentDidMount` or `componentDidUpdate` methods only.

akiran: instead of setting `slickList.style.height = height`  like this,  extract the necessary css properties in componentDidMount and save them in state.  and use that state to calculate style

As you said, we can have markup like this in main render function.
and getStyle functions shouldl be purely based on state
```
<div>
   <List style={this.getListStyle()}>
       <Track style={this.getTrackStyle()}>
            {this.props.children}
       </Track>
   </List>
   <LeftArrow />
  <RightArrow />
 <Dots />
<div>
```

vramana: I see. I will fix that.

PR #60, opened 2015-05-27
Add reactify as a dependency to use with browserify

### LINKED ISSUE #20: Reactify dependency missing from your package.json
Attempting to use your react-slick component and I'm getting a build error, that looks to be because you don't specify "reactify" as a dependency anywhere in your package.json even though you're using it as a browserify transform.  You must have it installed locally, but just forgot to commit that to the package.json?

```
>> Error: Cannot find module 'reactify' from './node_modules/react-slick'
Warning: Error running grunt-browserify. Use --force to continue.
```

vramana:
Fixes #20

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,8 @@
     "react-hot-loader": "^1.2.7",
     "react-responsive-mixin": "^0.3.1",
     "react-router": "^0.13.3",
-    "slick-carousel": "^1.3.15"
+    "slick-carousel": "^1.3.15",
+    "reactify": "^1.1.1"
   },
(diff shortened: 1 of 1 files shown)


akiran:
Putting it in  devDependency didn't work 
https://github.com/vramana/react-slick/blob/reactify/package.json#L47

Should it be in dependency list

vramana:
@akiran  I added it as dependency.

akiran:
We can provide the compiled files, so users don't have to use reactify/babel to use this library.

PR #61, opened 2015-05-27
Reworked prepublish

### LINKED ISSUE #20: Reactify dependency missing from your package.json

vramana:
Fixes #20


.gitignore
```diff
@@ -6,3 +6,4 @@ demos/*
 demos1
 TODO.md
 npm-debug.log
+lib/
```
akiran: We shouldn't put lib in .gitignore 
Files inside .gitignore will not be published on npm.

vramana: lib will not be ignored with as it is not in .npmignore

Issue #56, opened 2015-05-25
afterChange is calling the callback function before the transition.


vramana:
Steps to reproduce:
-  run `npm start` in project folder and open http://localhost:8000
- click on carousel of Center Mode

PR #62, opened 2015-05-27
Fixed afterChange behaviour and implemented beforeChange

vramana:


vramana:
Fixes #56

src/mixins/helpers.js
```diff
@@ -190,17 +190,27 @@ var helpers = {
...
+      this.setState(nextStateChanges);
+    };
+
+    var callback2 = () => {
+      this.setState(nextStateChanges);
+
+      if (this.props.afterChange) {
+        this.props.afterChange(currentSlide);
+      }
+    };
```
vramana: @akiran Instead of callback1 and callback2. Can this be written more nicely with a single callback?

akiran: addEndEventListener, and setTimeout are for same purpose. We should not give different callback to them.
setTimeout is backup for addEndEventerListener.

What if you call the afterChange function in callback and pass the same callback to both. Will it work ?

src/mixins/helpers.js
```diff
@@ -190,17 +190,27 @@ var helpers = {
...
+      swipeLeft: null
+    };
+
+    // var callback1 = () => {
+    //   this.setState(nextStateChanges);
+    // };
+
+    var callback2 = () => {
+      this.setState(nextStateChanges);
+
```
akiran: You can clean up this next PR.
Just call it callback

PR #66, opened 2015-06-06
Lazy load and Fade

vramana:
- [x] Lazy Load
- [x] Fade


src/inner-slider.jsx
```diff
@@ -19,11 +19,25 @@ export var InnerSlider = React.createClass({
   getDefaultProps: function () {
     return defaultProps;
   },
-  componentDidMount: function () {
-    // Hack for autoplay -- Inspect Later
+  componentWillMount: function () {
```
akiran: I think setting of mount in state should be in componentDidMount

vramana: Here the `setState` doesn't change anything so a rerender of VDOM is not required.

PR #68, opened 2015-06-11
fix logic when infinite = false

mathisonian:
The current version will still loop from one end of the slideshow to the other, even when the `infinite` property is set to `false`.

This pull request fixes that :boom:

### DIFF
--- a/src/mixins/helpers.js
+++ b/src/mixins/helpers.js
@@ -98,13 +98,17 @@ var helpers = {
 
     targetSlide = index;
     if (targetSlide < 0) {
-      if (this.state.slideCount % this.props.slidesToScroll !== 0) {
+      if(this.props.infinite === false) {
+        currentSlide = 0;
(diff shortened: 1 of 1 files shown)


vramana:
@mathisonian Which version did you observe the issue with ? Can you the settings with which I can reproduce this ?

vramana:
I think this will happen with `autoplay` set to `true`.

mathisonian:
Hey @vramana I was just using a basic setup like

```jsx
      <Slider arrows={false} infinite={false}>
        <MySliderPage />
        <MySliderPage />
        <MySliderPage />
      </Slider>
```

mathisonian:
I observed the behavior with `react-slick@0.5.0` from npm and the `master` branch on github

vramana:
Anyway thanks for the PR! I'll take a look at it locally then merge it.

PR #69, opened 2015-06-11
add support for 'unslick' responsive breakpoint setting

quarklemotion:
- porting this feature from the original slick carousel
- setting disables the slick carousel for the selected breakpoint (nested HTML not modified/slicked)
- responsive setting section of README updated with an example usage

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ var SimpleSlider = React.createClass({
 | focusOnSelect  | bool | | | 
 | infinite       | should the gallery wrap around it's contents | Yes |
 | initialSlide   | int | which item should be the first to be displayed | Yes (since pull req #17) | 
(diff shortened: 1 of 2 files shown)


vramana:
@quarklemotion Instead of setting `settings: unslick` better would be `settings: { unslick: true }` because the former way you have to add aditional logic when there are no breakpoints and the latter way you don't have to add any additional logic. What do you think ? 

Also don't update the dist files. They will updated at the time releasing. Don't bother to update them.

quarklemotion:
@vramana I was attempting to port over the feature from jquery slick carousel "as is" - and this is how you "unslick" with the jquery slick carousel.  See the settings example for the responsive demo at http://kenwheeler.github.io/slick .  If diverging this port a bit is OK, I'm happy to change the 'unslick' implementation in this PR to make it a 'regular' setting.

vramana:
@akiran What do you say?

akiran:
This looks good for me.  We can mostly follows slick's conventions and depart from it only when absolutely necessary.

@quarklemotion  Thanks for the pull request

PR #76, opened 2015-06-27
Add npmignore to cut down on package size

amsardesai:
If we add a `.npmignore` file, we can cut down on the number of files downloaded when users who use this package do `npm install`.


vramana:
@amsardesai Please merge the commits into one.

amsardesai:
@vramana done

vramana:
@amsardesai Thanks for your contribution!!

.npmignore
```diff
@@ -1 +1,17 @@
...
+bower_components
+.sass-cache
+build
+demos
+demos1
+TODO.md
+test
+testlib
+docs
+dist
```
akiran: It is better to keep the dist folder.
Some people may prefer to add those files in script tags instead of using browserify/webpack

amsardesai: Hmm, good point. I'll make another PR.

PR #81, opened 2015-07-17
Update to add support for IE9 transforms via `msTransform`

### LINKED ISSUE #80: No support for IE9 transforms
Noticed an issue where IE9 wasn't correctly changing which slide was selected because `msTransform` wasn't being set in `trackHelper.js`'s `getTrackCSS` function. Not sure if this lib intends to support IE9 or not but thought I'd open an issue anyway. (A `matchMedia` polyfill is required for `< IE 10` which leads me to believe this lib might just be 10+.)

Also, if you could provide instructions on building the project locally that'd be much appreciated, seems like `gulp dist` doesn't totally work quite as expected. It's Friday though, so that could explain it.

Thanks so much for this project, really great! :+1:

bdvis:
To resolve #80.

### DIFF
--- a/src/mixins/trackHelper.js
+++ b/src/mixins/trackHelper.js
@@ -27,7 +27,8 @@ export var getTrackCSS = function(spec) {
     WebkitTransform: 'translate3d(' + spec.left + 'px, 0px, 0px)',
     transform: 'translate3d(' + spec.left + 'px, 0px, 0px)',
     transition: '',
-    WebkitTransition: ''
+    WebkitTransition: '',
+    msTransform: 'translateX(' + spec.left + 'px)'
   };
 
   return style;


bdvis:
Let me know if that failed Travis build was my fault.

vramana:
@bdvis Don't you need also add the `msTransition` in `getTrackAnimateCSS` as well as `getTrackCSS` ?? Can you just check if all the demos work before I commit this?

The Travis failure is not your fault. I think @akiran turned on Travis testing for this repo but there is no .travis.yml in the repo. So you don't have to worry about it.

bdvis:
@vramana nope, because IE 9 doesn't support `transition`, so basically in IE9 you get no animation at all (it jumps immediately from one slide to the next). That was fine for our needs, but maybe this isn't the best overall solution?

vramana:
@bdvis Oh okay. You have to wait for @akiran to make the release or you use the github master for now.

bdvis:
Awesome. Thanks guys!

akiran:
@bdvis , @vramana  Released 0.6.2 with this change

bdvis:
:beers:

PR #95, opened 2015-08-05
Add slides into lazy load list based on the initial slide

sean-shilin:
lazyLoadedList should also consider the initialSlide config setting


vdv73rus:
When it would be merged in master?

akiran:
Fixed this in latest release

PR #96, opened 2015-08-05
Honors responsive breakpoints when calculating slides to show

mgmorcos:
Defers rendering _slider_ until breakpoints propagate through state, because _slide width_ is calculated once and for all at first render frame of _inner slider_ (`initialize()` in _helper mixin_).

### DIFF
--- a/src/slider.jsx
+++ b/src/slider.jsx
@@ -11,7 +11,8 @@ var Slider = React.createClass({
   mixins: [ResponsiveMixin],
   getInitialState: function () {
     return {
-      breakpoint: null
+      breakpoint: null,
+      isReady: !this.props.responsive,
     };
   },
   componentDidMount: function () {
@@ -27,7 +28,7 @@ var Slider = React.createClass({
           bQuery = json2mq({minWidth: breakpoints[index-1], maxWidth: breakpoint});
         }
         this.media(bQuery, () => {
-          this.setState({breakpoint: breakpoint});
+          this.setState({breakpoint: breakpoint, isReady: true});
         });
       });
 
@@ -35,11 +36,14 @@ var Slider = React.createClass({
       var query = json2mq({minWidth: breakpoints.slice(-1)[0]});
 
       this.media(query, () => {
-        this.setState({breakpoint: null});
+        this.setState({breakpoint: null, isReady: true});
       });
     }
   },
   render: function () {
+    if (!this.state.isReady) {
+      return null;
+    }
     var settings;
     var newProps;
     if (this.state.breakpoint) {


vramana:
@mgmorcos Can explain this a little bit using an example? I am not able to visualize properly.

mgmorcos:
```es6
 {
  slidesToShow: 4,
  slidesToScroll: 1,
  responsive: [
    {
      breakpoint: 1040,
      settings: {
        slidesToShow: 3,
        slidesToScroll: 1,
        infinite: true,
      }
    },
    {
      breakpoint: 850,
      settings: {
        slidesToShow: 2,
        slidesToScroll: 1,
        infinite: true,
      }
    },
    {
      breakpoint: 480,
      settings: {
        slidesToShow: 1,
        slidesToScroll: 1,
        infinite: true,
      }
    }
  ]
}
```

@vramana A slider with the above settings shows 4 slides on a small mobile screen, instead of 1 as configured, because the media queries that determine which responsive settings to use are async while the calculation of slide width is synchronous (in `initialize()` of helpers mixin).

This patch prevents premature calculation of slide width by delaying render of the slider and inner slider (which includes helpers mixin) until after the media queries are complete and breakpoints are set on state.

src/slider.jsx
```diff
@@ -35,11 +36,14 @@ var Slider = React.createClass({
...
 
       this.media(query, () => {
-        this.setState({breakpoint: null});
+        this.setState({breakpoint: null, isReady: true});
       });
     }
   },
   render: function () {
+    if (!this.state.isReady) {
+      return null;
```
iam4x: You have to return a valid component,`this.media` is async?

iam4x:
@vramana I ran into this issue too, the first render doesn't get the correct responsive settings. My fast workaround in my `render()` function:

```
let settings = {
      dots: true,
      useCSS: true,
      infinite: true,
      slidesToScroll: 4,
      adaptiveHeight: true,
      responsive: [
        { breakpoint: 9999, settings: { slidesToShow: 4, arrows: true } },
        { breakpoint: 1100, settings: { slidesToShow: 3, arrows: true } },
        { breakpoint: 900, settings: { slidesToShow: 2, arrows: true } },
        { breakpoint: 768, settings: { slidesToShow: 1, arrows: false } }
      ]
    };

    if (window) {
      const { innerWidth } = window;
      const { responsive } = settings;

      const correctSettings = responsive.reduce(
        (previous = { breakpoint: 0 }, current) =>
          previous.breakpoint > innerWidth &&
          current.breakpoint > innerWidth &&
          previous.breakpoint > current.breakpoint ?
            current : previous
      );

      settings = { ...settings, ...correctSettings.settings };
    }
```

I'm using server-side rendering too, so @mgmorcos solution don't work with me as I want to show content without client-side JavaScript enabled so waiting for async operations without DOM won't work.

vramana:
@iam4x I am bit busy. I'll look it into this in the weekend.

ChrisCinelli:
@iam4x  Do you have a PR ?

akiran:
Responsive option is working well in recent release.

This is no longer required

PR #103, opened 2015-08-19
IE8 Compatibility

### LINKED ISSUE #102: IE8 Compatibility?
don't work in IE8 totally.

sorrycc:
Close #102

### DIFF
--- a/src/dots.jsx
+++ b/src/dots.jsx
@@ -25,8 +25,9 @@ export var Dots = React.createClass({
       slidesToScroll: this.props.slidesToScroll
     });
 
-    var dots = Array.apply(null, {length: dotCount}).map((x, i) => {
+    var dots = [];
 
+    for (var i=0; i<dotCount; i++) {
       var className = classnames({
         'slick-active': (this.props.currentSlide === i * this.props.slidesToScroll)
       });
@@ -38,12 +39,12 @@ export var Dots = React.createClass({
         currentSlide: this.props.currentSlide
       };
 
-      return (
+      dots.push(
         <li key={i} className={className}>
           <button onClick={this.clickHandler.bind(this, dotOptions)}>{i}</button>
         </li>
       );
-    });
+    }
 
     return (
       <ul className={this.props.dotsClass} style={{display: 'block'}}>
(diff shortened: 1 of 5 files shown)


src/dots.jsx
```diff
@@ -25,8 +25,9 @@ export var Dots = React.createClass({
       slidesToScroll: this.props.slidesToScroll
     });
 
-    var dots = Array.apply(null, {length: dotCount}).map((x, i) => {
+    var dots = [];
 
+    for (var i=0; i<dotCount; i++) {
```
vramana: @sorrycc I don't see a need for this change. Please revert changes on this particular file. 

I'll look at this PR in the weekend. I am bit busy with my college work.

sorrycc: It's don't work in IE8 either.

nosir: Assume one has included es5-shim for IE8 supporting, Array.map should be safe. The problem is within `.apply`. It should be: `var dots = Array.apply(null, Array(dotCount)).map(...`
Tested with native IE8, it works.

src/dots.jsx
```diff
@@ -38,12 +39,12 @@ export var Dots = React.createClass({
         currentSlide: this.props.currentSlide
       };
 
-      return (
+      dots.push(
```
nosir: As above, there's no need to change logic here

src/dots.jsx
vramana: @nosir Thanks for your comment. My knowledge about ie8 is next to nothing. @sorrycc Is es5-shim an acceptable solution in your case? I really don't want to keep any for loops in my code unless it is required seriously.

nosir: @vramana To make ReactJS work with IE8, you have to include es5-shim anyway. Also I think it's not necessary to include the polyfill for addEventListener, it should be specified in README like ReactJS does in its doc.
http://facebook.github.io/react/docs/working-with-the-browser.html#browser-support-and-polyfills
I can send another PR for all these stuffs

nosir: Please refer to: https://github.com/akiran/react-slick/pull/109/
Thanks to @sorrycc for targeting the issue.

PR #109, opened 2015-08-21
Fix IE8 Compatibility

nosir:
This is based on https://github.com/akiran/react-slick/pull/109 from @sorrycc

### DIFF
--- a/src/mixins/trackHelper.js
+++ b/src/mixins/trackHelper.js
@@ -31,6 +31,15 @@ export var getTrackCSS = function(spec) {
     msTransform: 'translateX(' + spec.left + 'px)'
   };
 
+  // To detect IE8
+  //
+  // Do not use !window.addEventListener for browser testing here
+  // since the polyfill prototype will fix everything.
+  // Also this can avoid the detection being affected by any polyfill
+  // introduced out of the box.
+  // So here we use a native JS operation detection.
+  //
+  // Credit: http://stackoverflow.com/a/30617541/1849458
   if (!-[1,]) {
     style.marginLeft = spec.left + 'px';
   }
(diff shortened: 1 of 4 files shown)


src/mixins/trackHelper.js
```diff
@@ -31,6 +31,10 @@ export var getTrackCSS = function(spec) {
     msTransform: 'translateX(' + spec.left + 'px)'
   };
 
+  if (!-[1,]) {
```
vramana: @nosir Can you explain what is this??

nosir: You can't use window.addEventListener to test IE8 anymore, since the polyfill fixes everything. And we should assume any polyfill could possibly be included in any project. So here we use a native JS feature detection.

Credit: http://stackoverflow.com/a/30617541/1849458

vramana: Can you add a comment in source code as well?

nosir: Done

nosir: Please hold on for a sec, the current array in IE8 still has problem, I will need to push a fix.

nosir: @vramana ok, I've pushed the fix, maybe that's why @sorrycc went for `for` loop.
`map` doesn't iterate with empty array unless it's prefilled.

It might be related to this: https://github.com/es-shims/es5-shim/issues/190 I just added a quick fix for this case.

nosir:
Note: was checking this https://github.com/facebook/react/issues/3131 looks like any polyfill could break ReactJS feature testing. Closed the PR for now, I will have a look maybe go back to sorrycc's way.

PR #110, opened 2015-08-22
Fix IE8 Compatibility

nosir:
- Add matchMedia polyfill for README
- Fix addEventListener issue
- Fix empty array map issue

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,3 +97,6 @@ npm install
 npm start
 open http://localhost:8000
 ```
+
+### Polyfills for old IE support
+`matchMedia` support from [media-match](https://github.com/weblinc/media-match)
--- a/src/dots.jsx
+++ b/src/dots.jsx
@@ -25,7 +25,10 @@ export var Dots = React.createClass({
       slidesToScroll: this.props.slidesToScroll
     });
 
-    var dots = Array.apply(null, {length: dotCount}).map((x, i) => {
+    // Apply join & split to Array to pre-fill it for IE8
+    //
+    // Credit: http://stackoverflow.com/a/13735425/1849458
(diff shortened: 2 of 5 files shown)


vramana:
@nosir Have you tested this in IE8?

akiran:
Thanks to @nosir,  @sorrycc for trying to add support for IE8

@sorrycc Can you review this.
If it is fine, we will merge it and make a release soon

nosir:
@vramana yes, i've tested on native IE8

vramana:
Can you merge your commits into one? 

What about map? We have to put a notice to use es5-shim I guess.

nosir:
Merged to one.
map is working fine now, just filled the empty array first, please see the comment in dots.jsx
es5-shim is mentioned as a must thing for IE8 in ReactJS doc, otherwise the whole ReactJS doesn't load, so I guess it is not that necessary here.

One thing, I had problem setting up babel /webpack dev server on my home laptop, so I made change and tested within the distributed file in master, which should be the same. But if it's easy for you, would be nice if you guys can help generate the final js again and send me for testing. cheers.

akiran:
@nosir I merge the PR and generated dist files.
https://github.com/akiran/react-slick/tree/master/dist

I didn't have ie8 to test
If you can confirm, I will release it on npm

nosir:
Tested, all good. Thanks

nosir:
Thanks for updating, I didn't go for polyfill because ReactJS was differentiating the event usage by checking if `window.addEventListener` exists in dom. Not sure if they have changed this in ReactEventListener, will check later.

cytGitHub:
hellow , which version can support IE8 ?

PR #124, opened 2015-09-16
Making beforeChange working when fade=true

krasimir:
This PR fires `beforeChange` when `fade` is set to `true`. The call was missed in the `slideHandler` function. There are also two other small changes:

* the positioning of the dots in the demo page
* the failing tests


vramana:
@krasimir Please undo the changes on `dist/react-slick.js` and `dist/react-slick.min.js` They will be updated only in the release versions.

krasimir:
@vramana Ah right. I'll do that now.

krasimir:
@vramana done.

src/labelling.jsx
```diff
@@ -0,0 +1,40 @@
...
+  return slides;
+};
+
+
+export var Labelling = React.createClass({
+
+  render: function () {
+    var template, className, label;
+
+    if (this.props.labellingTemplate) {
```
vramana: Use ES6 destructuring. `const { labellingTemplate, slideCount, ... } = this.props;` 

Just use `const` all over the place. I'll do for the rest of the library later.

May be use labelTemplateFn / labelTemplateFunc instead of labellingTemplate. Instead of Labelling just use Label.

src/labelling.jsx
```diff
@@ -0,0 +1,40 @@
...
+    label = template(
+      this.props.currentSlide + 1, 
+      getSlidesCount({
+        slideCount: this.props.slideCount,
+        slidesToScroll: this.props.slidesToScroll
+      })
+    );
+
+    return (
+      <div className={ this.props.labellingClass ? this.props.labellingClass + ' ' + className : className }>
```
vramana: The conditional here should be moved into the `className` above.

src/labelling.jsx
```diff
@@ -0,0 +1,40 @@
+'use strict';
+
+import React from 'react';
+import classnames from 'classnames';
+
+var getSlidesCount = function (spec) {
```
vramana: Use seem to use this only at one place. Why not just inline the result?

src/labelling.jsx
krasimir: Thanks for the deedback @vramana. I'll address the issues and will prepare a new commit.

src/labelling.jsx
vramana: I would write this as 

```javascript
const getSlidesCount = ({ slideCount, slidesToScroll }) => Math.ceil(slideCount / slidesToScroll);
```

krasimir: That's a good idea. I'll do that. Thanks.

krasimir:
@vramana can you please check the latest commit and let me know what you think.

src/label.jsx
```diff
@@ -0,0 +1,35 @@
+'use strict';
+
+import React from 'react';
+import classnames from 'classnames';
+
+export var Label = React.createClass({
+
+  render: function () {
+    var template, className, label;
```
vramana: use `let` for `template` & `const` for the other two. Also declare where they are used.

src/label.jsx
```diff
@@ -0,0 +1,35 @@
...
+
+    const { labelTemplateFn, slideCount, slidesToScroll } = this.props;
+
+    if (labelTemplateFn) {
+      template = labelTemplateFn;
+    } else {
+      template = (currentSlide, slidesCount) => `${ currentSlide } / ${ slidesCount }`;
+    }
+
+    className = classnames(
```
vramana: I would do this as 

```
const className = classnames({
  'slick-label': true,
  [labelClass]: labelClass ? labelClass: ''
});
```

src/label.jsx
```diff
@@ -0,0 +1,35 @@
...
+      { 'slick-label': true },
+      this.props.labelClass ? this.props.labelClass : ''
+    );
+    label = template(
```
vramana: Nitpick
 
```javascript
const label = template(currentSlide + 1, Math.ceil(slideCount / slidesToScoll));
```

vramana:
@krasimir Comment on the PR diff instead of the comment. That was old comments are hidden after the revision.

vramana:
You forgot to exclude `dist/react-slick.min.js` from the PR.

akiran:
@krasimir, @vramana  labelling is not part of original slick API.

I think, we should solve this usecase in generic fashion by giving access to current slide state to the parent component of slider. (This is how original slick support this usecase)
This will let users create labels account to their requirement instead of using a default style provided by us

src/label.jsx
krasimir: Sorry, what you mean by *declare where they are used*. A comment maybe.

vramana:
@akiran I think this is a nice piece API. Please let us keep it. If we have problems with it we can discard it later.

docs/_style.scss
```diff
@@ -43,6 +43,20 @@ h3 {
 .slick-slider {
   margin: 30px auto 50px;
 }
+.slick-dots {
```
vramana: @krasimir How does the slider look with and without this rule?

krasimir:
@vramana the changes are done. `react-slick.min.js` is also reverted to its original form.
@akiran there is `labelTemplateFn` which is a prop of the component . It's a function that looks like that:
```js
labelTemplateFn: (current, all) => {
  return `You are looking ${ current } slide of ${ all } in total.`
}
```
So the users may provide a custom content of the Label.

docs/_style.scss
krasimir: Without it the dots are a little bit to the right side of the image:
image
With it they are centered:
image

krasimir: The left margin comes from the default browser behavior when we add `<ul>` tags.

vramana: Oops!! I never noticed that. Nice catch!

krasimir: I wouldn't see that without the Label placed at the bottom :)

akiran:
@krasimir  Nice work.

In my previous comment I am talking about generic API close to slick
Slick API has slickCurrentSlide property which returns the current slide. We should implement it.
```
<div>
    <Slider {...options} />
    <Label currentSlide={'current slide'} />
</div>
```
I think, this approach is more generic than passing an option to slick,
Making current slide accessible will be useful for other use cases as well and API will be short

krasimir:
@akiran I see. Indeed it will be much flexible. 

In general, I was a little bit surprised that there are no events dispatched by the component (let me know if I missed them). I was searching for something like `onSlideChanged`. This may be a little bit against the React policy of keeping the component's state for internal uses but will be useful for additional features like the label that I needed. For example:
```js
var onSlideChange = (current, all) => {
  // ... update the Label here.
}
<Slider {...options} onSlideChange={ onSlideChange }/>
```
Following this approach the component will know nothing about labelling. The dots may be extracted too.

vramana:
@krasimir You always add a getCurrentSlide method to the Slider component which returns the this.state.currentSlide. 
```javascript
  getCurrentSlide = () =>  {
    return this.state.currentSlide
  }
```
Then one can just attach a ref on top of the Slider component and can call 
`this.refs.slider.getCurrentSlide()`. It took me a little time to come up with this. But it seems nice a way of doings things.

A similar approach can be extended towards dots I think.  I am trying to do something similar [here](https://github.com/vramana/react-flex-slick)

krasimir:
There may be a problem with `getCurrentSlide`-ish functions. They represent part of the internal state and as such the data that they return changes. So if I need the change the value of my custom label then I should constantly execute `getCurrentSlide` which leads to a `setTimeout` in my code. I think the process should be more like `component -> push -> outside world` and not `outside world -> pull -> component`. Or in other words **react-slick: Oh, I just changed my current slide. Here is its index**.

akiran:
@krasimir I agree with you
For events and currentSlide, methods can be passed as props and slick should call them when necessary.

For example, if we pass slickCurrentSlide method, slick should call it with currentSlide argument when ever slide changes.

Method on parent component can be implemented like this
slickCurrentSlide: function (currentSlide) {
   // we can do anything we want here
  // setState on root component
  // or call flux actions etc
}

Similar approach can be used for event as well.

Let me know if your are interested in sending PR for this.
Or I will try to get to this in couple of days.

krasimir:
@akiran I'm fine preparing a PR but I don't know the component in details, so it will take me quite a lot of time to post the places where we need those calls.

akiran:
@krasimir  Ok. No problem. I will try to get it done soon.
Thanks for the ideas

krasimir:
@akiran you are welcome.

krasimir:
@akiran @vramana is there something else in this PR that needs updates. Are you willing to merge and release a new version. We are using a hard coded copy right now and will be nice to go back with npm installed one.

akiran:
@krasimir I think, we won't merge this because it deviates from slick API.

Once the current state accessible feature is implemented as discussed above, you can use that to implement labeling in you app

krasimir:
Thanks for letting me know @akiran.

akiran:
@krasimir  There are two options called beforeSlide and afterSlide.
If you pass these props, slick will cal l beforeSlide with currentSlide, nextSlide arguments and afterSlide with currentSlide argument.

This feature is already implemented. Fixed an issue with it in latest commit #edbc3766bd9ca1744e965ed107b68c728654ee45 

I am releasing the fix to npm now

Let me know if it works for you

krasimir:
Hey, that sounds really good. I'll try it today and will let you know.

krasimir:
Closing because `beforeChange` and `afterChange` are better ways of implementing labelling.

krasimir:
Ok this is weird. GitHub won't let me create another PR with appropriate description and not having all these messages above. Anyways, @akiran @vramana can you please check the final changes here. `beforeChange` didn't work when we have `fade: true`. The calling of the callback was missing in the piece that takes care for the fading (in `slideHandler` function.

krasimir:
The PR fixes also:
* the positioning of the dots in the demo page
* the failing tests

docs/demos.jsx
```diff
@@ -377,4 +377,4 @@ var App = React.createClass({
   }
 });
 
-module.exports = App;
+module.exports = App;
```
vramana: Please keep the new line here.

vramana:
@krasimir Just squash all you commit into one.

vramana:
Change name of you PR.

akiran:
@krasimir  Thanks for fix .

Please squash the commits to make it easy to refer in future.
I will merge and make a release

krasimir:
Yep, as soon as I go back to my pc :)

.npmignore
```diff
@@ -6,7 +6,6 @@ demos1
 TODO.md
 test
 testlib
-src
```
vramana: Why ?? Its not necessary.

krasimir: Ah ... let me bring it back. I removed it when I was testing the component locally.

akiran: Lets keep the src for now.  It is necessary to deploy docs.
I will clean it later

krasimir: Yep, fixed.

vramana:
remove react-slick.min.js

dist/react-slick.js
```diff
@@ -1106,7 +1102,7 @@ return /******/ (function(modules) { // webpackBootstrap
 	    }
 
 	    if (this.props.beforeChange) {
-	      this.props.beforeChange(currentSlide);
+	      this.props.beforeChange(this.state.currentSlide, currentSlide);
```
vramana: Hey, you shouldn't change this file

krasimir: That's really weird. I'm not sure why this change is visible here. I see that that's already merged in master https://github.com/akiran/react-slick/blob/master/dist/react-slick.js#L1104-L1106 

Maybe that comes because I had to merge with the master in this repo and sync my fork.

vramana: @krasimir You shouldn't use `master` branch for development

krasimir: @vramana I agree. The whole process was more like hacking the component. I didn't expect that will have to submit a PR here :)

vramana:
@krasimir since there are very few changes create another branch from this repos master and submit a PR again.

PR #125, opened 2015-09-16
replaced TransitionEnd callbacks with setTimeout

lPadier:
In React 0.14-rc1, CSSTransitionGroup changed the way they listened to the end of an animation from listening to the EndEvent, to setting a timeout prop as listening to the end event was not reliable enough.
https://facebook.github.io/react/blog/2015/09/10/react-v0.14-rc1.html
https://github.com/facebook/react/issues/1326

Moreover, on iOS, swiping on the slick during a scroll could sometimes block the component in an animating state.


vramana:
@lPadier Thanks for your PR!

vramana:
@akiran Can you merge this PR??

vramana:
@lPadier Actually, now that I think about it. Your PR is a bit incomplete. You should do clearTimeout of the setTimeout you are creating in componentWillUnmount. Otherwise, if you unmount mid animation it might throws errors. Please change this accordingly.

lPadier:
@vramana good catch, thanks.
I also created a commit that has this pull request and the folders created by `npm run pre-publish` committed:
https://github.com/lPadier/react-slick/commit/e00b8991995eb8327ed7023b487ac49eca7d3843
You can use it by updating the react-slick line in your package.json to 
```json
"react-slick": "lPadier/react-slick#e00b899",
```

vramana:
@lPadier Nice. Now this is good to go.

src/mixins/helpers.js
```diff
@@ -206,12 +205,12 @@ var helpers = {
...
       };
 
-      callback = () => {
+      this.animationEndCallback = () => {
         this.setState(nextStateChanges);
         if (this.props.afterChange) {
           this.props.afterChange(currentSlide);
         }
-        ReactTransitionEvents.removeEndEventListener(this.refs.track.getDOMNode(), callback);
+        delete this.animationEndCallback;
```
akiran: @lPadier I think you have to use clearTimeout  and not delete callback

iam4x: He didn't named `this.animationEndCallback = window.setTimeout` so there's no need of `clearTimeout` in this case.

src/inner-slider.jsx
```diff
@@ -50,6 +50,9 @@ export var InnerSlider = React.createClass({
     if (this.resizeListener) {
       this.resizeListener.remove();
     }
+    if (this.animationEndCallback) {
+      clearTimeout(this.animationEndCallback);
```
iam4x: This won't work.

src/mixins/helpers.js
```diff
@@ -108,14 +107,14 @@ var helpers = {
         if (this.props.afterChange) {
           this.props.afterChange(currentSlide);
         }
-        ReactTransitionEvents.removeEndEventListener(this.refs.track.getDOMNode().children[currentSlide], callback);
+        delete this.animationEndCallback;
```
lPadier: I delete the reference to the animation timer as it is no longer necessary.
There is no need to to a clearTimeout as the timeout has already ended.

iam4x:
Looks good to me! 

@akiran Any ETA or want some help with new contributors?

akiran:
I will all the review all the open PR's and merge them by this weekend.

akiran:
Incorporated this fix in latest release

PR #133, opened 2015-09-26
Slider did not update current slide to initialSlide prop on update

dptoot:
External control methods were left out of the port due to their anti-pattern nature of parent updating state.  There have been several requests for the ability to update the current slide externally and the solution posted has always been to update the initialSlide prop.  Unfortunately the current update method does not use this to update the current index.  I have added a slideHandler method using the initialSlide index as the argument to facilitate an animated update of the slider.

I have also added a demo as it is a recurring question in issues.

### DIFF
--- a/docs/demos.jsx
+++ b/docs/demos.jsx
@@ -356,6 +356,54 @@ var Fade = React.createClass({
   }
 });
 
+var ExternalControl = React.createClass({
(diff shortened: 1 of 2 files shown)


akiran:
Now you can use slickGoTo to change the slide after mounting.

PR #144, opened 2015-10-13
adds support for fractional `slidesToShow`

gbrassey:
allows fractional values in the `slidesToShow` option

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # react-slick
 Carousel component built with React. It is a react port of [slick carousel](http://kenwheeler.github.io/slick/)
 
-For React 0.13 or below version support, please stick with react-slick@0.8
+For React 0.12 or below version support, please stick with react-slick@0.7.2
 
 # Road to 1.0
 react-slick is under active development now.


robcolburn:
`slick-carousel` seems to allow "peeking" while still using integers as slidesToShow.

See Examples:
* Variable Width
* Center Mode
* Slider Syncing

I'm not really advocating one approach or another, but this project seeks parity with `slick-carousel`.
https://github.com/akiran/react-slick#road-to-10

gbrassey:
Yes. I realize now that I was having trouble because I was using React elements as direct children to the `Slider` and not applying the `style` prop (which passes the slide width). I think this pull request can be closed, although I wonder if there is anyway to protect against this type of developer error?

robcolburn:
React PropTypes doesn't distinguish integers vs numbers :/. But, we could add a custom validation in propTypes

https://facebook.github.io/react/docs/reusable-components.html

gbrassey:
I was thinking more along the lines of ensuring that the width of the slides is set by react-slick, even if the direct children of the `Slider` element are other react components. 




The problem arose because my slide components did not set the style that is passed to them by react-slick in the props (ie. `<slideComponent style={this.props.style}>`). I solved the problem by wrapping the `slideComponent` in `div`'s. I see `React.cloneWithProps` is being used to render the slides but it was not setting the width without explicitly declaring it in jsx.

PR #147, opened 2015-10-14
fix for latest react warnings/deprecations

rkoberg:
Here are fixes for the using this module with the latest react.


src/mixins/helpers.js
```diff
@@ -8,9 +9,9 @@ import assign from 'object-assign';
 var helpers = {
   initialize: function (props) {
     var slideCount = React.Children.count(props.children);
-    var listWidth = this.getWidth(this.refs.list.getDOMNode());
-    var trackWidth = this.getWidth(this.refs.track.getDOMNode());
-    var slideWidth = this.getWidth(this.getDOMNode())/props.slidesToShow;
+    var listWidth = this.getWidth(this.refs.list);
+    var trackWidth = this.getWidth(ReactDom.findDOMNode(this.refs.track));
```
iam4x: You can use directly `this.refs.tracks`. Because DOMNode is already bound to refs (see: https://facebook.github.io/react/blog/2015/10/07/react-v0.14.html#dom-node-refs)

src/track.jsx
```diff
@@ -68,7 +67,7 @@ var renderSlides = (spec) => {
     }
 
     var childStyle = getSlideStyle(assign({}, spec, {index: index}));
-    slides.push(cloneWithProps(child, {
```
iam4x: Care, I'm not sure this is wanted behaviour. Each slide should have his own props, see: https://facebook.github.io/react/docs/top-level-api.html#react.cloneelement

rkoberg: The warning I saw had advice to do it with `slides.push(React.cloneElement(child, {`, which removed the warning.

rkoberg: So if each slide can't share props with the ability to override props, the docs you point me to say it should use `React.addons.cloneWithProps`. Is that what you mean? The slideshow does work with what I have it the pull request.

src/mixins/helpers.js
rkoberg: Hmmm... When using it directly, I get:
```
Uncaught TypeError: elem.getBoundingClientRect is not a function
```

akiran:
@rkoberg  Thanks for PR.  
Since 0.14 got released recently, many people will still be using 0.12, 0.13
I want to support atleast two earlier versions.

Can you make changes such that it work for both 0.14 and 0.12.
May be we can have if statements
if (React.version < 0.14.0) {
   // code with getDOMNode
} else {
 // code without getDOMNode
}

Once react 0.15 is released, we can stop supporting 0.12.

rkoberg:
Can someone who is already set up with those react's give it a go? Otherwise it would probably need to wait till the weekend.

robcolburn:
@akiran - How can I test the library in alt versions of React?  The tests seem to be empty, so can you describe your manual process?  I'll can help write some up tests.

Most of this should be backwards compat in React 0.13, and I've got things I can test, but React 0.12 would be difficult for me to verify.  I'm also curious how Node would handle the referencing.  AFAIK, I think even requiring `react-dom` would break in React 0.12.

Right now I'm thinking…
```js
const minR14 = require('semver-compare')('0.14.0', React.version) >= 0;
const findDOMNode = minR14 ? require('react-dom').findDOMNode : ref => ref.getDOMNode();
const cloneElement = minR14 ? React.cloneElement : require('react/lib/cloneWithProps');
```

akiran:
@robcolburn  Thanks Rob . That will be great.
Can you submit PR with semver changes you proposed.

iam4x:
@robcolburn Why support old releases of React? People should move on, if they don't they should stick to old version of react-slick.

Just bump the version with breaking changes, semver will take care of the rest :+1:

robcolburn:
@iam4x,
Thanks, but it's not my call. The project author, @akiran, said that he wanted to maintain company for 2 versions of React https://github.com/akiran/react-slick/pull/147#issuecomment-148104507 .

IMO: I see value in supporting 1 version back of React while the community is migrating. After that just bump major version of the module to alert the breaking change. But, the project author/maintainer  can decide how much growing pains they'd like to take on

iam4x:
Sorry my bad, my comment was for @akiran :+1:

rkoberg:
I would bet that people who are using react are pretty nimble with versions and probably stay up to date.

dmitry:
@rkoberg not really, depends on the project. We have medium-sized project, and we don't have enough resources at the moment to upgrade. Plus not all the dependent projects are moved to the 0.14.

akiran:
If react-slick has all the features and if it stable, we can release a new version supporting only the latest version.  But few more features need to be added. 

So we can support two previous versions until this library is stable.

karlnp:
@rkoberg @dmitry I keep close to whatever I need to work stably. Sometimes that means that a library with a fixed bug that I'm running into fixed it partially by updating React to 14, which is the problem I'm facing at the moment.

rkoberg:
I added support for the "slick-current" class.

rkoberg:
Any word on getting this merged? I am going to add the lib dir to my fork so I can install from the github repo.

matsilva:
Any update on this getting merged? Those console warnings are nasty. Every other react dependency I have does not play well with anything <0.14 since those packages have all already migrated to support 0.14 in their latest releases.

zawilliams:
+1 on an update on getting this merged

akiran:
This PR also included build files in dist, lib. So I couldn't merge it.

#153 is clean with out build files. So I merge #153 and release as 0.9.0-rc1 for react 0.14 support

akiran:
I will decide whether to support previous versions or not soon and release 0.9.0
Until then you can use 0.9.0-rc1

PR #149, opened 2015-10-20
Add support to provide custom classes for slides

victor-castellanos:
- Add tests for rendering with default class
- Add tests for rendering slides with custom class
- Modify props to expect null custom class

### DIFF
--- a/src/track.jsx
+++ b/src/track.jsx
@@ -26,12 +26,15 @@ var getSlideClasses = (spec) => {
   } else {
     slickActive = (spec.currentSlide <= index) && (index < spec.currentSlide + spec.slidesToShow);
   }
-  return classnames({
+  var { slidesClass } = spec;
+  var classNames = classnames({
     'slick-slide': true,
     'slick-active': slickActive,
     'slick-center': slickCenter,
     'slick-cloned': slickCloned
   });
+
+  return slidesClass ? `${classNames} ${slidesClass}` : classNames;
 };
 
 var getSlideStyle = function (spec) {
--- a/test/slick.test.js
+++ b/test/slick.test.js
@@ -1,7 +1,51 @@
-describe('Slick test', function() {
-  
-  it('should render slides with dots', function() {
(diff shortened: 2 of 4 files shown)


akiran:
@victor-castellanos  Thanks Victor. I will review it today.

victor-castellanos:
@akiran My pleasure man. I got a couple more ideas coming your way :+1:

src/track.jsx
```diff
@@ -26,12 +26,15 @@ var getSlideClasses = (spec) => {
...
+  var { slidesClass } = spec;
+  var classNames = classnames({
     'slick-slide': true,
     'slick-active': slickActive,
     'slick-center': slickCenter,
     'slick-cloned': slickCloned
   });
+
+  return slidesClass ? `${classNames} ${slidesClass}` : classNames;
 };
```
akiran: @victor-castellanos  I want the api of react-slick to be close to api of [slick](http://kenwheeler.github.io/slick/)
Since slick api didn't have slidesClass option we have to find another way to solve this usecase.

one way would be add custom styles to .slick-slide class.

test/slick.test.js
```diff
@@ -1,7 +1,51 @@
...
\ No newline at end of file
+
+  context('when rendering with custom slide class', () => {
+    let slides;
+
+    before(() => {
+      const slidesClass = 'customClass';
+      const component = renderIntoDocument(
+        <Slider {...props} slidesClass={slidesClass}>
+          <div><p>Random text</p></div>
```
akiran: You can  add the custom class in slides

For example
```
<div className='customClass'><p>Random text</p></div>
```
React slick will add remaining classes while keeping customClass
```
<div className='slick-slide slick-active customClass'><p>Random text</p></div>
```

akiran:
Closing this because this usecase can be solved without adding additional options

akiran:
@victor-castellanos  Thanks for PR

PR #151, opened 2015-10-20
Update initial slide on new thumbnail event

Andrewlee5889:
Not sure if I'm using this wrong. But this logic seems to make sense to us! Please let me know if this is accurate. Thanks!

When user selects a new thumbnail in a string of pictures, set currentslide to equal to initial slide when there's a new prop.initialSlide


akiran:
It could be misleading to change initialSlide.
Ideally it should used for setting the slide when initially mounted.

We can implement another prop "slickGoTo" to change the slides after mounting.
slickGoTo is also a part of [slick-carousel](http://kenwheeler.github.io/slick/) api

akiran:
@Andrewlee5889 Would you like to send a new PR with slickGoTo implementation.
Logic is same as this PR, but using a different prop name

Andrewlee5889:
Yessir! I'll work on it and make a new PR. Thanks!

Andrewlee5889:
@akiran Thanks for the feedback! My concern is that, my PR doesn't change the initialSlide. All it does is that IF the initialSlide prop changes, update currentSlide so that it reflects the change.

Let me know if that makes sense! If not I will create the new PR with slickGoTo tonight! Thank you

akiran:
I think, slickGoTo will be better for changing slides, initialSlide to specify the initial slide.

Andrewlee5889:
@akiran I updated the prop to be slickGoTo. Is this what you meant? Sorry for the trouble but thanks again!

PR #152, opened 2015-10-20
Fix a bug where autoplaying carousel would lock indefinitely when bro…

mll:
…wser window is resized during automatic slide change.


scriptype:
:+1:

mll:
@akiran

akiran:
Fixed this in react-slick@0.11

chadhahn:
This seems broken (again?) in 0.12.2. What was the fix for 0.11 if it wasn't this PR from @mll?

PR #154, opened 2015-10-23
Added ability to set the current slide

m4n3z40:
Hey, i've been using your awesome lib for some time now in my company, and it's been really useful, but now I needed this feature that it didn`t support it yet, and I've got it to work, check it out and see if it works for you. For now i'm changing my package dependency to my fork because a need this now.

Thanks!


akiran:
@m4n3z40 Now we can set the current slide with prop "slickGoTo'  implemented in #151
Let me know if it meets your use case

PR #155, opened 2015-10-23
Add some basic testing.

robcolburn:
There is testing infrastructure, but not yet any tests running.
This makes changes scary, as we worry about breaking things.  This is
some very basic testing of rendering and tearing down the examples.

Ideas to expand on this:
* Simulate Clicks and other expected user interactions.
* Test common patterns people actually use.
* Test alt versions of React that we want to support.


akiran:
@robcolburn Great start.

Issue #159, opened 2015-10-28
Slider childrens classNames are not merged anymore in 0.9.0-rc1


websilone:
I know 0.9.0-rc1 is still under development but here is a little issue.

Having className on slider children elements (which will become the .slick-slide), it seems these classNames are not merged anymore with the new ones.

From what I saw in the changes between 0.8 and 0.9, in track.jsx, line 70, 0.9 is using React.cloneElement where 0.8 was using cloneWithProps.
So it's probably an intern change in these methods but it would be nice to find a way to keep these classNames on the child elements.

Thanks.

akiran:
yes, I will look into this

websilone:
if this can help, here is what I came up with.

In track.jsx, in the renderSlides method, line 69 : 

var slickClasses = getSlideClasses(assign({index: index}, spec));
    var cssClasses;

    if (child.props.className) {
        cssClasses = classnames(slickClasses, child.props.className);
    }
    else {
        cssClasses = slickClasses;
    }

    slides.push(React.cloneElement(child, {
      key: index,
      'data-index': index,
      className: cssClasses,
      style: Object.assign({}, child.props.style || {}, childStyle)
    }));

It is actually due to React.cloneElement not merging styles and classNames anymore for us.

dfala:
@websilone +1 -- classNames not merged

akiran:
It is fixed with #161

PR #160, opened 2015-10-29
Fix the fact that childrens classNames are not merged anymore

websilone:
Following the issue #159 , here is the fix I suggested in the comments.

Just checked if the child.props.className exists, if so, merging these props with the calculated classes from the getSlideClasses method.


websilone:
My bad... created the fix branch from master instead of 0.14...
Closing this PR.

akiran:
Thanks @websilone 
Can you open a new PR branching from 0.14

websilone:
Yes I already did that... #161

PR #161, opened 2015-10-29
Merging classnames fix (from the right 0.14 branch this time...)

websilone:
Following the issue #159 , here is the fix I suggested in the comments.

Just checked if the child.props.className exists, if so, merging these props with the calculated classes from the getSlideClasses method.


montanehamilton:
Just merged these changes into my custom build and they fix the issues I was having, love to see these merged soon.

PR #163, opened 2015-10-30
Replaced Object.assign for assign, and fixed resize to work on IE8

msangui:


montanehamilton:
This should be merged as well later versions of IE are also effected by this. Manually patched my version and it seems to do the trick.

akiran:
@msangui  There is merge conflict. Can you resolve it

msangui:
@akiran done!

PR #164, opened 2015-11-01
Update webpack.config.dist.js

montanehamilton:
Leaving react-dom in the distribution files opens us up to two versions of react running...


montanehamilton:
Added updated dist files.

Issue #177, opened 2015-11-24
child's className prop lost when clone slides in infinite mode


soulwu:
[L90-L108](/akiran/react-slick/blob/0.9.2/src/track.jsx#L90-L108)
The lines above clone preCloneSlides and postCloneSlides when using {infinite: true}, but I found it just ignore original child's className prop.

is it an issue?
if so, I could fix it and push a pull request.:smile:

PR #178, opened 2015-11-24
fix currentSlide prop and children's className prop

soulwu:
related to #177


soulwu:
add submit to fix #179

Issue #180, opened 2015-11-28
option to temporarily disable forward/back arrows and dots


tnrich:
I'd like to be able to, based on a prop passed in to the component, disable the forward arrow, the back arrow, or both at the same time. Does anyone know if there is there a way to do that?

Thanks!

tnrich:
Doesn't look like the functionality I want exists yet, so I'll try PRing it. It looks like the test suite isn't totally functional yet, so I hope I'm not breaking anyone else's work...

akiran:
You can use nextArrow and prevArrow for this usecase

PR #182, opened 2015-11-30
adding ability to disable dots and previous/next arrows via props

tnrich:


tnrich:
Connects to #180

akiran:
I don't want add new options which are not part of slick API

PR #183, opened 2015-12-01
Fix adaptive height lag

tnrich:


tnrich:
Connects to #184

akiran:
@tnrich I don't want to extra options to disable dots since it is not a part of original slick api.

Can you separate fix for adaptive height bug  and submit a new PR

gkilian:
I assume this is here to somewhat help with this?
https://github.com/akiran/react-slick/issues/429

PR #192, opened 2015-12-09
Fix issue when clicking prev/next faster than currentSlide is re-calc…

mortentangen:
`targetSlide` is not always simply the next increment, it could be an infinite positive/negative number if the user clicks fast enough on the prev/next arrows (this is easily done if there exists more than a few slides, since the animation seems to go much faster)

Result after going bananas on the next arrow without this fix:
image


Cunha44:
+1

akiran:
@mathisonian  This fix didn't solve the problem completely .

I checkout you code and clicked on next button of first demo very fast. 
Still I see white screen

mortentangen:
Yes, it doesn't seem like the cloning of slides is fast enough to avoid seeing the white screen, but with this fix at least you won't get stuck with a white screen when you finish clicking.

akiran:
ok.  I will try to fix the white screen issue.

dijonkitchen:
Any update on this?

PR #194, opened 2015-12-09
Make the classnames applied to the `prevArrow` and `nextArrow` configurable

jonathanp:


akiran:
@jonathanp We needn't add additional options to pass arrow classnames to keep API simple.
We can use nextArrow and prevArrow props to pass arrow component with required classnames.

We should remove this className prop to prevent className of custom Arrow component form getting overwritten
https://github.com/akiran/react-slick/blob/master/src/arrows.jsx#L25

Issue #197, opened 2015-12-14
Autoplay ignores rtl


ranbena:
When setting prop `rtl: true`, it's expected that the autoplay would go *left to right*, but instead goes *right to left*.

akiran:
This issue it fixed

PR #198, opened 2015-12-14
Autoplay ignores rtl #197

ranbena:
A simple condition on `this.props.rtl` to choose between adding or subtracting to determine the next slide.


akiran:
Thanks @ranbena  for PR

Issue #169, opened 2015-11-09
Freeze autoplay carousel in Demo page


kuy:
I didn't dig this issue why it occurred, but I'd like to report and note steps to reproduce.

*Steps to reproduce*

1. Open [Demos](http://webrafter.com/opensource/react-slick) page in Chrome
1. Scroll down to Autoplay carousel
1. Keep Demos tab, and switch to other Chrome's tab
1. Wait 5 minutes
1. Back to Demos tab
1. Autoplay carousel is stopped

laxplaer:
+1

donaldparker:
+1

jessy1092:
Hi all, I create PR #199 for it. :)

Cunha44:
+1

akiran:
duplicate of #202

PR #199, opened 2015-12-17
Add autoplayPause props to support pause autoPlay

jessy1092:
If autoplayPause is true, it would stop slide animation but keep autoPlay timer.

I think it is more flexible for user to define their autoplay behavior.

Relative issue #169


akiran:
This optional is not available in original slick http://kenwheeler.github.io/slick/

I think, we should fix the issue without adding new option to keep the API consistent with slick.

PR #216, opened 2016-01-15
add "react-dom"

avantcontra:
add "react-dom" in package.json.


akiran:
We haven't yet droped the support for react 0.13.  
Adding read-dom will force users to use react 0.14

PR #222, opened 2016-01-22
:bug: Fix cloned slides doesn't copy className prop

VincentBel:
Before:  
image

After fixed:

image

### DIFF
--- a/src/track.jsx
+++ b/src/track.jsx
@@ -4,7 +4,7 @@ import React from 'react';
 import assign from 'object-assign';
 import classnames from 'classnames';
 
-var getSlideClasses = (spec) => {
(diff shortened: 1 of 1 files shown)


FakhruddinAbdi:
@VincentBel  I have a same problem and i have tried your solution, but didn't work for me, 
Could you please update your solution

VincentBel:
Hi, @FakhruddinAbdi. It seems that this has been fixed by 6f5b551. And this has already been discussed in #177 
And it seems that @akiran hasn't released a new version after fixed, so you may have to try the codes on  master branch to find out whether it works for you.   
Closed by 6f5b551.

PR #229, opened 2016-02-02
Add option to debounce window resize functions

dcocchia:
If there are a large number of carousels on a single page, especially if those carousels have complex children components, the page will lag very badly while resizing. Sometimes taking many seconds to finish on even a modern computer/browser. This adds an option to debounce the resizing function so adjustments are only made once the user is finished resizing their window.


akiran:
Need to add debounce functionality without adding additional option

Check the original slick implementation
https://github.com/kenwheeler/slick/blob/master/slick/slick.js#L1773-L1785

Would you like to open new PR with this change ?

akiran:
opened https://github.com/akiran/react-slick/issues/233 issue to track this

PR #236, opened 2016-02-15
Fixing issue where current slide would go past the slidecount

fmolina24:
Seems like if autoplay is on and infinite is false the current slide will continue to increment. This causes a problem because the arrows will reflect that currentSlide meaning you will have to click the arrow until you reach the last slide.


akiran:
Fixed autoplay issue in react-slick@0.11.0
Can you test it

PR #237, opened 2016-02-16
Modify PrevArrow disabled state calculation

marsbomber:
When calculating disabled state on the PrevArrow, we can rely on this.props.currentSlide to determine if we should disable PrevArrow or not.

This fixes the following problem.

We have slidesToShow: 2, slideCount: 4. When we slide right to show item 3 and 4 on screen, we dynamically delete item 3 and item 4, where we will have slidesToShow: 2, slideCount: 3, the PrevArrow is disabled and we cannot scroll back to show item 1 and 2


akiran:
I think, best way to deal with this scenario is move to slide 1by calling slickGoTo(0) before deleting 3rd and 4th slides.

(this.props.slideCount <= this.props.slidesToShow) condition is essential to disable prev button when slide count is less than slides to show

PR #239, opened 2016-02-16
Set z-index to move up current slide to top

kuy:
This PR fixes an issue on Fade type.
When slides contain clickable contents, I can't click a content of the current slide. What I can click is always **last slide**.

**Issue**: Can't click a content of current slide
**Fix**: Set `z-index` to slides

Please notice that the solution was brought from original slick library.

### DIFF
--- a/src/track.jsx
+++ b/src/track.jsx
@@ -42,7 +42,13 @@ var getSlideStyle = function (spec) {
   if (spec.fade) {
     style.position = 'relative';
     style.left = -spec.index * spec.slideWidth;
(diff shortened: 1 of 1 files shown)


kuy:
Hi @akiran, would you give your comment on this?
I'll fix if something wrong.

escapedcat:
This can be done with css only by using the `slick-active` or/and  `slick-center` css-classes, no? That's what I did. Looks ok so far.

akiran:
Yes, CSS is the best way to add z-index.

PR #242, opened 2016-02-19
dotsHumanReadable prop

welbornio:
This adds the ability for you to toggle the dots to "human readable" (prop name suggestions?). This would mean that the content inside the dots is 1, 2, 3, etc, rather than 0, 1, 2.


akiran:
This feature has to be implemeted using customPagging option
https://github.com/kenwheeler/slick/blob/master/slick/slick.js#L54-L56

welbornio:
Thanks for pointing that out. I'll work on a new PR.

akiran:
Thanks @welbornio 
I opened an issue to track to this feature https://github.com/akiran/react-slick/issues/243

PR #244, opened 2016-02-22
matching default functionality to slick

welbornio:
You took my attention to these lines in my last PR: https://github.com/kenwheeler/slick/blob/master/slick/slick.js#L54-L56

and I noticed this time around that the default for the dots content is actually { i + 1 } instead of { i }. So this basic change actually matches the default dot content for slick.

Here's a screenshot of the content from the first dot on slicks webpage.

screen shot 2016-02-21 at 7 05 06 pm

This does, however, affect backwards compatibility. How would you like to proceed?

Also, I'm working on the customPaging option in a separate branch.

### DIFF
--- a/src/dots.jsx
+++ b/src/dots.jsx
@@ -43,7 +43,7 @@ export var Dots = React.createClass({
 
       return (
         <li key={i} className={className}>
-          <button onClick={this.clickHandler.bind(this, dotOptions)}>{i}</button>
+          <button onClick={this.clickHandler.bind(this, dotOptions)}>{i + 1}</button>
         </li>
       );
     });


welbornio:
I'm going to close this until I can figure out why the build is failing

welbornio:
@akiran Any insight into why this is failing? I also pulled current master and npm run test failed locally with the same error.

Thomas0c:
@welbornio Since the error thrown did not provide much explanation (for me at least), I saw that the used version of PhantomJS was outdated (`1.9`). Latest version is `2.1.3` and updating to that seems to have fixed the problem.

akiran:
@welbornio Ignore test failures for now.  I will fix the test setup soon and start writing tests.

In future, you can submit PR with code fix along with test .

akiran:
It should not effect much, because that number will be replaced by dot if css and fonts are available.

welbornio:
@akiran do you mean as far as backward compatibility goes? I know that it's not a big change, but I have a use case on a project where we use the dots as numbered boxes, and we need the correct number.

akiran:
Yes,  I mean its not a big problem for backward compatibility.
I will merge it today

welbornio:
Awesome! Glad to hear it.

PR #247, opened 2016-02-26
Change this.props.children.length to use React.Children.count() instead.

psynewave:
> Warning message "Iterable.length has been deprecated, use iterable.size or iterable.count()" when using ImmutableJS list as items

When using a immutable collection with react-slick you will receive the above error message. This pr fixes the issue by changing the collection measurement from length the to use the React.Children.Count.

P.S. This is my very first pull request so please let me know if I need to change anything.


akiran:
@psynewave  Thanks for PR

Issue #254, opened 2016-03-04
Issue with stacking slides before react-slick is initialized…


chemoish:
https://github.com/akiran/react-slick/blob/master/src/inner-slider.jsx#L80

`slick-initialized` is hard coded and not set by any events.

![screen shot 2016-03-03 at 8 14 59 pm](https://cloud.githubusercontent.com/assets/549319/13517870/a648dcee-e17c-11e5-88e8-c524eb1e9551.png)

How can you prevent stacking slides? Thoughts?

_In kenwheeler/slick, `slick-initialized` is an event… so you can prevent FOUC._

akiran:
Render slick slider only when you need.
slick-initialized won't be necessary in react

chemoish:
I need it immediately? or am I missing something?

I solved this the FOUC issue by…

```jsx
componentDidUpdate() {
    this._slider_container.classList.add('initialized');
}

render() {
    return (
        <div className="slider-container" ref={c => (this._slider_container = c)}>...</div>
    );
}
```

```css
// styles

.slider-container
{
    display: none;

    .initialized &
    {
        display: block;
    }
}
```

akiran:
You don't need to do that.  check this demo

https://jsfiddle.net/kirana/20bumb4g/

chemoish:
I am thinking that it has to do with server side rendering. Since the DOM and images are already delivered.

akiran:
I am not sure if slider can support server-side rendering well. 
It does some math to compute dimensions of slides and slide-track based on container width, which is known only after rendering on browser

Randy808:
Not sure if this is related, but why is the first slide stacked in this example and is there a way to resolve it? :
https://jsfiddle.net/vddg2baj/2/

anshuPurohit:
@chemoish I solve FOUC by using this technique:
1) on server side pass infinite: false and pass only 1 item to slick component
2) on client side pass  infinite: true and pass all the items to slick component

This way slick HTML received in the browser will have only 1 item to show and when slick component is initialized on the client side it will render all the items.

rajatjasuja23:
@anshuPurohit How are you implementing server side code for infinite :false to the slick component ?

dcurletti:
@anshuPurohit although your solution does improve the user experience, two things still occur:
the rest of the images pop into view
if your slider is center to "centerMode" there will be an even larger user disturbance as the image moves to the center and the other images are rendered.

laveesingh:
[...]

PR #256, opened 2016-03-04
Fix `slick-initialized` to fire when the component finishes mounting …

chemoish:
…(Only fires on the client). This will prevent a FOUC when doing SSR.

NOTE: https://github.com/akiran/react-slick/blob/master/src/initial-state.js#L27 seems like dead code.

### DIFF
--- a/src/inner-slider.jsx
+++ b/src/inner-slider.jsx
@@ -40,6 +40,8 @@ export var InnerSlider = React.createClass({
     }
   },
   componentDidMount: function componentDidMount() {
+    this._slickSlider.classList.add('slick-initialized');
+
     // Hack for autoplay -- Inspect Later
     this.initialize(this.props);
     this.adaptHeight();
@@ -77,7 +79,7 @@ export var InnerSlider = React.createClass({
     this.update(this.props);
   },
   render: function () {
-    var className = classnames('slick-initialized', 'slick-slider', this.props.className);
+    var className = classnames('slick-slider', this.props.className);
 
     var trackProps = {
       fade: this.props.fade,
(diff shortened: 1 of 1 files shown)


chemoish:
Addresses https://github.com/akiran/react-slick/issues/254.

chikathreesix:
Works great! Hiding the component is definitely better than FOUC. Ideally I'd like to prepare the space for the carousel and show a loading indicator in there.

chemoish:
@akiran thoughts?

chemoish:
@chikathreesix, for me `slick-initialized` is just wrong as it currently stands and should be fixed. #imo

chikathreesix:
@chemoish I see your point. I have also tried a different approach after I commented here #288

akiran:
Closing this because add classes with classList.add is not a very react friendly
I might consider merging #288

chemoish:
Do you have documentation that indicates your statement is the case?

In my experience this is the most performant way of managing visibility. Alternatively, you can have a `getInitialState` and a `setState` that manages this information—it will cause a double render, but the virtual DOM should make the hit trivial.

Either solution is better than #288—due in part to the complexity of the solution, but also the current logic is incorrect (The slider isn't initialized until the component finishes mounting).

chikathreesix:
The reason why I created #288 is, this solution hide the whole carousel so it still causes a layout change after the code runs. The logic of #288 is not complicated, it just hides non current banners. What makes it not really beautiful is handling duplicated banners placed on both left and right. In my opinion, the carousel should take another approach not to generate these backup banners but this is a different topic so I have just dealt with it in a not really beautiful way.

PR #260, opened 2016-03-14
add slick-arrow class to prev/next buttons as per slick-carousel

rcambrj:
These classes are on slick-carousel prev/next buttons, so they should be on react-slick ones, right?

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "main": "./lib",
   "scripts": {
     "start": "gulp server",
+    "postinstall": "babel ./src --out-dir ./lib",
     "prepublish": "babel ./src --out-dir ./lib && gulp dist",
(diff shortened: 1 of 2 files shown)


akiran:
I couldn't merge it because you changed prepublish to postinstall.

But I added your arrows class name change

rcambrj:
Ah! my bad - I didn't mean to put that commit into the pull request, so thanks for sorting it in parallel.

I made that change because when I referenced a fork of this repository directly (instead of via npm) the module installation wouldn't take place correctly (it was a while back so I'm unsure exactly why). You might want to consider an alternative fix if ` "postinstall": "babel ./src --out-dir ./lib",` isn't acceptable.

akiran:
No problem.  Thanks for PR

PR #268, opened 2016-03-23
If the slider receives new props with fewer children than before and …

### LINKED ISSUE #267: Slider becomes empty and unnavigable when the number of slides is reduced
To repro:

* set up a slider that displays 4 slides at a time, with 5 slides
* navigate to the second page (1 slide will be in view)
* re-render the slider with only 4 slides

Expected:
* the user is moved back to the first page and can see the first 4 slides.
* the dots and navigation buttons disappear or become disabled, because there is only one page now.

Actual:
* the fifth slide disappears and the user has an empty slider.
* the dots and navigation buttons disappear or become disabled, so you can't use the buttons to go back to the first page.
* however you can get back to the first page by dragging with the mouse.

Here's a fiddle that demonstrates the problem: https://jsfiddle.net/89wqwvvx/3/

adamshone:
…the current slide index is now greater than the number of children, reset the current slide index to the start of the final page.

Fixes #267

### DIFF
--- a/src/inner-slider.jsx
+++ b/src/inner-slider.jsx
@@ -60,6 +60,12 @@ export var InnerSlider = React.createClass({
     }
   },
   componentWillReceiveProps: function(nextProps) {
+    if (this.state.currentSlide >= nextProps.children.length) {
+      this.setState({
+        currentSlide: nextProps.children.length - nextProps.slidesToShow
+      });
+    }
+
     if (this.props.slickGoTo != nextProps.slickGoTo) {
       this.changeSlide({
           message: 'index',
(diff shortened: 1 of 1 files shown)


HriBB:
This piece of code causes problems for me.

This is my render function.
```
return (
  <ReactSlick {...slickOptions}>
    <div>
      <Slide
        imageSize={imageSize}
        location={location}
      />
    </div>
    {location.sectors.map((sector, index) => (
      <div key={sector.id}>
        <Slide
          key={sector.id}
          imageSize={imageSize}
          location={location}
          sector={sector}
        />
      </div>
    ))}
  </ReactSlick>
)
```

In my case `nextProps.children.length` is 2, when it should be 12. This is because of the way react handles elements ...

```
Array[2]
  0: Object
  1: Array[11]
  length: 2
```

We need to use `React.Children.count(nextProps.children)` and detect children length change.

PR #272, opened 2016-04-02
Improve slide movement trigger logic

jonathanewerner:
We're using react-slick for [http://wish.technology/en/features](our website). It's been quite nice, so thanks for your work on this library!
There have been some minor issues, though. One of them is the slide movement trigger logic. As you might see when using our website on a touch device (i'll switch the npm version with this PR though soon, so it might not be visible in 2 days), the sliders are way to "trigger-happy", meaning that they activate although i just want to scroll down the page. 
I tried to grok the code and optimize the behaviour so that this trigger-happyness is not an issue anymore.
I would assume that this could be a general benefit for the library, which is why i PR'd this and not just left it in a private fork.

Interestingly, the relevant check had been implemented *after* the `setState` that modifies the tracks left position, making it effectively a noop if i'm not mistaken.

So basically, i changed the code order so that the check comes *before* said `setState` and tuned the constant to be more "ergonomic" + explained in code what makes it ergonomic (imo). :)

What do you think?


src/mixins/event-handlers.js
```diff
@@ -3,6 +3,52 @@ import {getTrackCSS, getTrackLeft, getTrackAnimateCSS} from './trackHelper';
 import helpers from './helpers';
 import assign from 'object-assign';
 
+/**
+ *
+ * Calculate if horizontal track swiping should occur.
+
+ * A typical "vertical" thumb swipe to navigate the page can have a degree offset
```
escapedcat: Is there a source link for this? Nevermind if you did this yourself.

daniel-rabe:
we have the same problem; if scrolled in a slide the slide shouldn´t move. your solution helps, but has one problem; if the user swipes the "correct" way and continues to swipe in the "incorrect" way (e.g. to left and then down) the slider stucks in between two slides

Issue #279, opened 2016-04-08
Skip key generation if slides already have key property


ViLPy:
Hi,
There is an issue with auto-generated key property which appears in case of manipulating slides before removing them. Because key property is re-generated every time from zero, React keeps all applied styles/css classes/etc for removed tile, because for React you only remove last key.

See https://jsfiddle.net/4cp9keq6/2/ 
On click slide opacity set to 0 before removing. If you click slide A (key = 0), after removing React translate all applied styles to slide B, which became slide with key = 0 after removing A.

akiran:
#281 PR is merged.

bryanbraun:
I found this issue, but it looks like the issue has come back. I went ahead and opened a new issue (with a new example demo) over here: https://github.com/akiran/react-slick/issues/2005

PR #281, opened 2016-04-08
Keep original keys for slides

ViLPy:
Fix for #279 
Skip auto-generated keys for slides if key already exist.


jackfengji:
This will cause the duplicate key problem when using infinite option

PR #288, opened 2016-04-13
Fix FOUC for Server Side Rendering

chikathreesix:
Another approach to #256.
This is to prevent FOUC when you render it on the server side.

Currently it shows all elements vertically so my approach is hiding non current elements until it gets sideWidth so that only the current element is displayed on the first load.
Also passing a key of an element to getSlideStyle in order not to display left and right edge elements.

### DIFF
--- a/src/track.jsx
+++ b/src/track.jsx
@@ -37,6 +37,7 @@ var getSlideStyle = function (spec) {
 
   if (spec.variableWidth === undefined || spec.variableWidth === false) {
     style.width = spec.slideWidth;
+    style.display = (!spec.slideWidth && (spec.currentSlide !== spec.index || spec.key < 0 || spec.children.length <= spec.key)) ? 'none' : null;
   }
 
   if (spec.fade) {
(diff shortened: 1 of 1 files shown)


Der1906:
I have the same problem, @akiran can you merge this please!?

varmil:
I have the same problem, so I hope this pr is merged soon :) 👍

vishwanatharondekar:
Bumping this up

vishwanatharondekar:
Just a observation. This doesnt take `slidesToShow` into consideration and always ends up showing only one slide irrespective of whatever value set for `slidesToShow `

src/track.jsx
```diff
@@ -37,6 +37,7 @@ var getSlideStyle = function (spec) {
 
   if (spec.variableWidth === undefined || spec.variableWidth === false) {
     style.width = spec.slideWidth;
+    style.display = (!spec.slideWidth && (spec.currentSlide !== spec.index || spec.key < 0 || spec.children.length <= spec.key)) ? 'none' : null;
```
vishwanatharondekar: This should somehow take `slidesToShow` into consideration.

PR #292, opened 2016-04-18
Fix initialSlide issue with lazyLoad enabled

### LINKED ISSUE #246: Initial silde will not be displayed when lazyLoad mode enabled
Initial silde will not be displayed when lazyLoad mode enabled, and set `lazyLoad: false` slide will be displayed correctly. It seem some weird behavior in lazyLoad mode. Below is my setting when encountered this issue:

``` js
const settings = {
      className: 'center',
      infinite: false,
      speed: 500,
      lazyLoad: true,
      slidesToShow: 3,
      centerMode: true,
      centerPadding: '60px',
};
```

Roulette:
Fixes #246


akiran:
Fixed this in latest release

Issue #295, opened 2016-04-19
Settings variation and slickGoTo


linuxenko:
It seems like after state change when i'm trying open slide with `slickGoTo`, slider does not work anymore (does not handle swipe events and i cannot change slide by dragging/swiping it).

With `dots: true` settings option `slickGoTo` works fine . But it makes DOM for dots, that not so cool.

Thank you

dougmacklin:
+1. I just updated to react@15.0 and react-slick@0.12.0 and slickGoTo is no longer working for me as well, it results in a blank slider.

HriBB:
Try setting `waitForAnimate: false`

akiran:
slickGoTo is now change to method instead of prop.
Check this example https://github.com/akiran/react-slick/blob/master/examples/SlickGoTo.js

open a new issue if you are still facing issues with slickGoTo

Issue #278, opened 2016-04-07
The slider has less slides won't display if switch from a slider has more slides


wuyiqun0718:
Hi there,

I was trying to use a dropdown menu to change the displaying slides. Option 1 has a few slides, option 2 only has 1 or 2 slides so it can't slide at all.  So when I select the option 1 first, and slide it to the right end. Then switch to option2, then the slides won't displayed at all and the next/prev buttons is disables.

I hope it's described clearly, this is the best I can do. Here's the jsfiddle:

https://jsfiddle.net/20bumb4g/231/

Thanks!

akiran:
Fixed in master. Will make a npm release soon

PR #298, opened 2016-04-19
Fix breakage when re-rendering with different children

### LINKED ISSUE #297: Slider can become unusable when altering the number of children
When re-rending the slider due to a change in the number of children (slides), certain situations can cause the slider to incorrectly it position itself far to the right of the window (off-screen).

The most common way I can reproduce this issue is by starting with a large list, with the current slide at least half-way through the list, then re-rendering with fewer children.

I have reproduced this issue in a [jsfiddle](https://jsfiddle.net/chrisknepper/sg9tn203/1/) and I have created a [short video](http://chrisknepper.com/misc/react-slick-index-error.mp4) demonstrating the problem.

This is likely the same problem causing #267, #278, and perhaps #295.

chrisknepper:
This addresses the slider disappearing and becoming unusable, relating to changing the children (slides). It fixes #297, and could also be the solution for #267, #278, and #295.

### DIFF
--- a/src/inner-slider.jsx
+++ b/src/inner-slider.jsx
@@ -61,7 +61,7 @@ export var InnerSlider = React.createClass({
   },
   componentWillReceiveProps: function(nextProps) {
     if (this.props.slickGoTo !== nextProps.slickGoTo
-      || this.props.children[0].key !== nextProps.children[0].key) {
+      || (this.props.children && this.props.children[0].key !== nextProps.children[0].key)) {
       this.changeSlide({
           message: 'index',
           index: nextProps.slickGoTo,
--- a/src/inner-slider.jsx
+++ b/src/inner-slider.jsx
@@ -60,11 +60,12 @@ export var InnerSlider = React.createClass({
     }
   },
   componentWillReceiveProps: function(nextProps) {
-    if (this.props.slickGoTo != nextProps.slickGoTo) {
+    if (this.props.slickGoTo !== nextProps.slickGoTo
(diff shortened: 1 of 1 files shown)


src/inner-slider.jsx
```diff
@@ -60,7 +60,8 @@ export var InnerSlider = React.createClass({
     }
   },
   componentWillReceiveProps: function(nextProps) {
-    if (this.props.slickGoTo != nextProps.slickGoTo) {
+    if (this.props.slickGoTo !== nextProps.slickGoTo
+      || this.props.children[0].key !== nextProps.children[0].key) {
```
akiran: @chrisknepper  Why are we checking only for children[0] key ?

chrisknepper:
@akiran It appears to be an efficient (fast) and reliable way to check if we should indeed call `this.changeSlide` instead of `this.update`. I can't think of a situation where this logic would be faulty, though another (slightly slower) solution would be to loop through `this.props.children` and `nextProps.children` in parallel, breaking out of the loop when a difference in keys is found.

We could also compare the lengths of `this.props.children` and `nextProps.children`, but we can't solely rely on length to determine what should be called in `componentWillReceiveProps`.

I did notice an error with my change if the slider is created with no children, so I've added an additional check for that.

dougmacklin:
Any chance we could merge this PR and republish to NPM? I heavily rely on `slickGoTo` and since updating to react@15.0 this functionality has been broken :( . Thanks again for all the hard work on this plugin!

akiran:
Fixed #297  in master. Will make a npm release soon

PR #309, opened 2016-05-06
Adds the class slick-displayed to all dots that have its slides being displayed

rafaelteixeira:
Useful to style the dots where not only the active one has a highlight.


akiran:
Sorry, this cannot be added as it is not present original slick API

This functionality can be achieved with custom class names you pass to slides and you can compute which which slides are shown with the help of currentSlide value given by beforeChange/afterChange methods

PR #333, opened 2016-06-13
Replace boolean values in eslint config with integers as per the docs

yn5:
Atom complains:

```
Configuration for rule "global-strict" is invalid:
Severity should be one of the following: 0 = off, 1 = warn, 2 = error (you passed 'true').
```


yn5:
The reason the travis check has failed is because of the bug fixed by: https://github.com/akiran/react-slick/pull/332

PR #335, opened 2016-06-13
Expose previousSlide and nextSlide functions

yn5:


psimyn:
do you have a use-case in mind for this? Specifically for using refs

and can you replace this PR with one without the auto generated files

thanks!

akiran:
Merge #409 which has similar fix

Issue #357, opened 2016-07-13
Bug with variableWidth and centerMode


bratva:
Hello!

I have slider with this options

```
var options = {
        infinite: false,
        variableWidth: true,
        centerMode: true
}
```

When i swipe last slide to left or first slide to right, the slider throw error `Cannot read property 'offsetWidth' of undefined`; in `getTrackLeft`

Demo https://jsfiddle.net/t1ovuve8/

chrishinrichs:
+1

nluck:
+1

maximusnikulin:
+1

roman-yakobnyuk:
Same here, it also occurs without infinite mode enabled, with just center mode.

chrishinrichs:
Might want to give https://github.com/godaddy/react-img-carousel a try

pjjanak:
@akiran - Has there been any movement on this? I'm experiencing a similar issue with variableMode = true and setting an intialSlide. Namely, the initial slide isn't brought into frame. Forcing an update after the component mounts makes it a little better, but the initial slide is still a little bit off.

It seems like you closed #325 which describes my issue marking it as a duplicate of this.

laveesingh:
Similar issues have been fixed in previous releases, please try the latest version. If the issue still persists for you, please create a codesandbox demo and reopen it.

josephdonnelly:
Unfortunately this issue still exists:
https://codesandbox.io/s/7m0qow8l9j

laveesingh:
The referenced commit should do it.

PR #359, opened 2016-07-14
fix Bug with variableWidth and not infinite

bratva:
https://github.com/akiran/react-slick/issues/357


akiran:
fixed in #526

bratva:
Thank you

PR #362, opened 2016-07-16
Unique component key for dynamic rendering

piq9117:
In track.jsx, `getKey()`, added `: child.key + fallbackKey;`


akiran:
@piq9117  I think, it is not a good idea to use index as key. It is better to generate unique key with uuid or some other method and pass unique keys  to the slides

akiran:
I don't see any benefit in adding index to the key already passed to slide.

piq9117:
```
Warning: flattenChildren(...): Encountered two children with the same key, `cloned0`. Child keys must be unique; when two children share a key, only the first child will be used.
    in div (created by Track)
    in Track (created by InnerSlider)
    in div (created by InnerSlider)
    in div (created by InnerSlider)
    in InnerSlider (created by Slider)
    in Slider (created by SliderWindow)
    in div (created by SliderWindow)
    in SliderWindow (created by CreateReport)
    in div (created by CreateReport)
    in CreateReport
    in main (created by MainLayout)
    in div (created by MainLayout)
    in MainLayout (created by Wrapper)
    in div (created by Wrapper)
    in Wrapper
```
Thank you for the review. I get this error when I dynamically add elements to the slider. What is the solution for this problem?

akiran:
Can you give a code snippet to replicate this issue ?

piq9117:
```
export default class SliderWindow extends Component {
    render () {
        const {photos} = this.props;
        const settings = {
            dots: true,
            infinite: true,
            speed: 500,
            slidesToShow: 1,
            slidesToScroll: 1
        }
        const images = _.map(photos, (p, i) =>
            // pretty much just gets the src from the parent
            <div key={i} style={style.imgContainer}><img key={i} src={p} style={style.img}/></div>
        );
        return (
            <div style={style.container}>
                <Slider {...settings}>
                    {images}
                </Slider>
            </div>
        );
    }
}
```
This is my child component where the Slider resides.

akiran:
I tried this  https://jsfiddle.net/20bumb4g/898/  
couldn't replicate your issue.

Can you replicate your issue with above jsfiddle and give me a link to debug

piq9117:
Somehow this works perfectly fine [https://jsfiddle.net/Piq9117/gon6295z/10/](https://jsfiddle.net/Piq9117/gon6295z/10/)
but it always shows an error in our app, maybe it's something else. Thank you for catering to my concerns. I'll get back to you once I get to the root of this problem. Again, thank you.

escobar022:
It seems that on the fiddle it works as it is production and therefore not throwing dev issues, i believe. I have tried this as well and am having the same issue. 
What I believe is happening is  that a key from slick-slider is being set to the dynamically created key, when it already has one, therefore causing issues. I could be wrong.

here is my snippet. I'm doing it a bit different, same issue.

`
var SDCWidget = React.createClass({
	render: function sdc_widget() {
		var settings = {
			dots          : true,
			infinite      : true,
			speed         : 500,
			slidesToShow  : 4,
			slidesToScroll: 1,
			buttons       : true,
		};

		var slides = data.slides.map(function (slideNode, index) {
			return (
				<div className="top-slider__item" key={index}>
					<Slide slide_info={slideNode} key={index} />
				</div>
			)
		});

		return (
			<div>
				<ModuleTitle module_title={data.module_title} />
				<div className="module-container">
					<Slider {...settings}>
						{slides}
					</Slider>
				</div>
			</div>
		);
	}
});

if (document.getElementById('sdc_widget')) {
	var data = {
		type        : 'sdc_widget',
		module_title: 'Most Popular',
		slides      : [
			{
				id  : 1,
				img : 'http://placekitten.com/g/300/200',
				desc: 'Cat does not know react'
			},
			{
				id  : 2,
				img : 'http://placekitten.com/g/300/300',
				desc: 'This Cat Knows React'
			}
		]
	};

	ReactDOM.render(
		<SDCWidget data={data} />,
		document.getElementById('sdc_widget')
	);
}
`

akiran:
#398 fixes this issue

ryanbagwell:
0.12.3 fixed the issue for me.

jessestuart:
Can confirm that upgrading to `0.12.3` fixed for me as well.

PR #367, opened 2016-07-20
pass through 'slickActive' prop to slide

blairanderson:
Looking for a cleaner way to know if the slide is being actively rendered or not. 

this would be helpful. 

Used `slickActive` instead of `rendered` or `active` because i think they're a bit generic.  

Helps us know when cards are seen or not.


akiran:
If you are using a react component for Slide, you can use 
```
this.props.className.indexOf('slick-active') > -1
```
to know if slide is active.

This logic need not be part of this library

PR #368, opened 2016-07-20
update afterChange to named parameters

blairanderson:
This callback would be much more useful if it included more data. 

things i'd like to see:

- was it manually rotated, or auto-rotated?
- which direction did it go?
- which slides are currently visible?(taking responsive into account)


akiran:
@blairanderson We can't merge this because of API change.
It will break existing users code

PR #377, opened 2016-07-24
Vertical

Josh-ES:
This build of the project adds working 'vertical' and 'verticalSwiping' properties. This consists of several changes when the 'vertical' property is set to true:

- The height CSS property of the '.slick-list' element is set manually.
- The height of the track is now set based on the number of children and the heights of a child (variableHeight or similar is not yet supported), and the width is not set in this case.

When 'verticalSwiping' is set to true:

- Vertical swiping is now detected correctly by the swipe event handlers, including swipeDirection. This includes the swipe length, min swipe and swipe direction being calculated correctly for vertical swiping. Horizontal swiping is ignored.


Josh-ES:
Added some extra commits to the PR to bring the base of my modifications into line with the main repos.

PR #378, opened 2016-07-25
SwipeToSlide

Josh-ES:
This is a basic implementation of the 'swipeToSlide' property, including readme and demo. A similar pull request for this feature already exists, but I've implemented the feature myself with a view to making the 'swipeToSlide' property compatible with the vertical mode and vertical swiping, something that is not present in the original slick carousel library. The following briefly details the changes made:

- .idea added to the gitignore for Webstorm (unrelated, can be removed).
- Description of 'swipeToSlide' added to the readme.
- SwipeToSlide demo added to the docs.
- 'build' script added to the package.json file that only includes babel transpilation (again unrelated, can be removed).
- getSlideCount, checkNavigable and getNavigableIndexes methods from the original library reimplemented.
- These methods used to provide swipeToSlide functionality in the swipeEnd method. If swipeToSlide is set to false the behaviour here is no different, and the number of slides moved is still the value in the 'slidesToScroll' property. Otherwise, the methods mentioned determine the number of slides to scroll based on the swipe event.


Josh-ES:
Added some extra commits to the PR to bring the base of my modifications into line with the main repos.

PR #384, opened 2016-07-29
Reset animating state when window onresize.

bergkampho:
Resize the screen when the slider is animating. The state `animating` keep in true and doesn't go to false after the animation transistion end. It causes `slideHandler` function always return. Thus, the slider stop working.


akiran:
@bergkampho Added this logic in resize event handler instead update, because update can be used in other cases.

PR #396, opened 2016-08-08
Added centerPadding feature

renddiyeh:


src/inner-slider.jsx
```diff
@@ -129,11 +130,28 @@ export var InnerSlider = React.createClass({
       nextArrow = (<NextArrow {...arrowProps} />);
     }
 
+    var centerPaddingStyle = null;
+
+    if (this.props.vertical === false) {
+      if (this.props.centerMode === true) {
+        centerPaddingStyle = {
+          padding: ('0px ' + this.props.centerPadding)
```
umesh-velir: would be great if this could handle top/bottom paddings as well, not just left/right.

PR #397, opened 2016-08-08
Make dots clickable optional witch by default is true

qawsedstation:
This PR will make the dots not clickable if we want, using this parameter dotsClickable.
By default is set to true so is not changing the current behaviour just make the component more flexible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,7 @@ var SimpleSlider = React.createClass({
 | centerPadding  | | | |
 | cssEase        | | | |
 | dots           | bool | Should we show the dots at the bottom of the gallery | Yes |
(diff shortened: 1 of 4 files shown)


akiran:
You can implements this feature by using dotsClass prop without adding extra option to package.

You can pass dotsClass as 'slick-dots disabled'

qawsedstation:
Hi there. I try that but is not working is still clickable..

davidglbr:
@qawsedstation 'slick-dots disabled' didn't work or me as well. Dots are still clickable and respond to clicks. 

@akiran would you consider accepting this commit?

PR #399, opened 2016-08-10
fix: swipeMove and swipEnd preventDefault event

chenpx976:


akiran:
@chenpx976  What issue does this fix ?

chenpx976:
I use this in the IOS webview
need to prevent native swip events or can not properly use

Issue #337, opened 2016-06-17
Slider completely stops working when re-rendering multiple times


mkralla11:
After rendering multiple times, even if no settings were changes for the react-slick slider, the slider completely stops working. No button clicks, no swipes work, and the slider is completely stopped. Here is the problem occuring:

https://jsfiddle.net/20bumb4g/727/

To reproduce, click the next and/or previous buttons until you notice that nothing is working anymore. All the code is doing is re-rendering the given component about every second.

Expected when rerendering:
The react-slider show continue working, including all buttons.

Actual Result:
The react-slider stops sliding and no buttons work anymore.

maxfriedmann:
I get the same result when I update a component while the slide animation is being done. In one of my projects I have some translated text on a sliding box, when I change the language while it is sliding, it stops sliding and the buttons don't work anymore. 

Maybe this is the same error?

Your Fiddle doesn't work anymore unfortunately...

mkralla11:
Do you mean the fiddle doesn't load or you can't actually use the slider in the fiddle? Anyways, I'm pretty sure you're having the same issue, as soon as you setState a few times on any parent component or the given slider wrapper component itself, all slider functionality breaks: no sliding, no buttons work. I recently switched to https://github.com/kenwheeler/nuka-carousel but that slider has it's own set of issues...I can't recommend it.

vemperor:
Yep, same here in chrome/safary (no problem in ff). It looks like callbacks "disappear" after rerender while scrolling and `animating` don't go to `false`. Don't know if that's expected behavior. Any ideas?

vemperor:
also, jsfiddle: https://jsfiddle.net/2gybogg4/1/

bergkampho:
I found the same issues while resizing the screen when the slide is animating.
The state `animating` keep in `true` and doesn't go to `false` after resize. It causes `slideHandler` always return.

pelhage:
I am having the same issue. When I update the children or re-render the slide stops working altogether

grossmail1:
Honestly, I just went to the standard slick implementation with Jquery. It felt dirty and wrong and was definitely a little tougher to implement. But it is far less buggy.

pelhage:
I have confirmed that the slideshow will break if it is re-rendered during an animation: https://jsfiddle.net/20bumb4g/1027/

If you click the "Trigger re-render button" you'll see that the slideshow breaks and is no longer functional

dcocchia:
[...]

PR #407, opened 2016-08-11
Fix issue where animation transition events do not get triggered.

dcocchia:
This should fix many issues where the `track` DOM node's transition events do not fire on transitionend, most likely because state changes during animation causes a re-render.


dcocchia:
A fi for https://github.com/akiran/react-slick/issues/337

blairanderson:
@dcocchia as per contributing guidelines, remove changes to `/dist`

dcocchia:
Will open in a new PR

dcocchia:
Moved to: https://github.com/akiran/react-slick/pull/408

PR #408, opened 2016-08-11
Fix issue where animation transition events do not get triggered

dcocchia:
Fix for https://github.com/akiran/react-slick/issues/337


akiran:
Replaced ReactTransitionEvents with setTimout completely and released 0.13.0 version.
Please check if your issues are fixed with latest version

PR #416, opened 2016-08-16
Fix two edge case issues.

zhulongzheng:
1. https://github.com/akiran/react-slick/commit/ed7447fbce84754116b334fc013769897ab5cb75 This commit fixed the back scrolling animation when there's only one slide and we set `inifinite: true`.
2. https://github.com/akiran/react-slick/commit/7634fa54b31a8bc8b979e4d0c0d3fbf37ceea0a8 This commit fixed the update of `currentSlide` when the number of slides may change frequently.

### DIFF
--- a/src/mixins/helpers.js
+++ b/src/mixins/helpers.js
@@ -43,6 +43,8 @@ var helpers = {
     var trackWidth = this.getWidth(ReactDOM.findDOMNode(this.refs.track));
     var slideWidth = this.getWidth(ReactDOM.findDOMNode(this))/props.slidesToShow;
 
+    var currentSlide = props.rtl ? slideCount - 1 - props.initialSlide : props.initialSlide;
+
     // pause slider if autoplay is set to false
     if(!props.autoplay)
       this.pause();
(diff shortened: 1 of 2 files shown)


src/mixins/helpers.js
```diff
@@ -43,6 +43,8 @@ var helpers = {
     var trackWidth = this.getWidth(ReactDOM.findDOMNode(this.refs.track));
     var slideWidth = this.getWidth(ReactDOM.findDOMNode(this))/props.slidesToShow;
 
+    var currentSlide = props.rtl ? slideCount - 1 - props.initialSlide : props.initialSlide;
```
akiran: Why do you want to change currentSlide during update ?

May be need we need to check and update currentSlide only if slideCount < currentSlide

zhulongzheng: > Why do you want to change currentSlide during update ?

Suppose we have a uploader and slider, the slider will update it's slides when we upload  items or remove items in the uploader, especially when we remove items.

> May be need we need to check and update currentSlide only if slideCount < currentSlide

Yeah, maybe your strategy is more conservative than mine. But it's up to you.

akiran: yes,  Issues like #297are happing when slide count change dynamically. I will fix it with the above check I mentioned.

Please open separate PR's when your are fixing multiple issues to make it easy to review

PR #425, opened 2016-08-19
Only scroll to the last visible item when a set is uneven and infinite = false

ryanbagwell:
Presently, when ```infinite = false```, and the carousel has an uneven set, it will slide to the end of the number of slides to show, even if the carousel doesn't have enough items to fill the space. Example below:

screen shot 2016-08-19 at 3 13 11 pm

The correct behavior (IMO) should be the same as if infinite were ```true```  - scroll until the end of the visible item, like this:

screen shot 2016-08-19 at 3 17 54 pm

This PR does the following: 

1. adds an example of an uneven set that is not infinite, and;
2. sets the offset so that it will only scroll to the last visible item.

### DIFF
--- a/docs/demos.jsx
+++ b/docs/demos.jsx
@@ -7,7 +7,8 @@ import SimpleSlider from '../examples/SimpleSlider'
 import SlideChangeHooks from '../examples/SlideChangeHooks'
 import MultipleItems from '../examples/MultipleItems'
 import Responsive from '../examples/Responsive'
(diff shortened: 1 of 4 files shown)


ryanbagwell:
@akiran thank you for merging. Can you also do a quick version bump and publish to npm when you have a moment?

stevemao:
@akiran @ryanbagwell is there a way to turn it off?

PR #430, opened 2016-08-22
bug if num of carousel items is equal to num of items displayed

puntubabu:
bug replication: https://jsfiddle.net/6pqv6g2o/ 

fixed bug where if `slidesToScroll == slidesToShow` AND is `infinite == true` AND `centerMode == true`, then track offset is not calculated correctly


akiran:
Couldn't merge this because of merge conflicts, but incorporated this change

Thanks for the PR

PR #431, opened 2016-08-23
slickGoTo method

akiran:


greenimpala:
Why deprecate this? Using pure props for data flow in React components is preferred practice.

PR #445, opened 2016-08-29
Fixes Swipe too far in infinite == false did not reset to previous position

louisptremblay:
Hi,

When the slider has infinite mode == false and we slide too far, the slider doesn't refresh the position of the slides on swipe end, so they are stuck in the middle instead of being moved back to the original position. The fix is that even if the slide number doesn't change, we should still update the state to refresh the position.

Before PR:
beforepr_image

After PR:
afterpr_image

### DIFF
--- a/src/mixins/helpers.js
+++ b/src/mixins/helpers.js
@@ -156,11 +156,6 @@ var helpers = {
       currentSlide = targetSlide;
     }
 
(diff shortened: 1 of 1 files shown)


leonaves:
Please merge this @akiran, such a simple fix and it's driving me mad.

leonaves:
ur a G m8

PR #449, opened 2016-08-31
Ref callbacks

Josh-ES:
Changed all refs attributes to use callbacks instead of strings, the use of which is considered legacy. See here for more details: https://facebook.github.io/react/docs/more-about-refs.html.


akiran:
@Josh-ES Thanks for the PR
I will review and merge soon

PR #453, opened 2016-09-01
Ref callbacks

Josh-ES:
Changed all refs attributes to use callbacks instead of strings, the use of which is considered legacy. See here for more details: https://facebook.github.io/react/docs/more-about-refs.html.

I've got a whole lot of extra features implemented on my fork of this repo, particularly swipe to slide and vertical, for a larger project I'm working on. Had to rebase this branch to keep up to date with main repo, didn't quite realise it would close the PR. Will be more careful in the future!!!


akiran:
@Josh-ES You can re-open PR's for swipe to slide and vertical.
I will review this week and merge.

Josh-ES:
I took my own fork of the code in a very different direction to the main repo, using decorators and ES6 classes in place of mixins and React.createClass for example, before implementing those changes. That's why I ended up closing the old PRs!

I could reimplement those features to bring new feature branches more into line with the current repo, not sure how you want me to proceed?

akiran:
I think it is better if you could open new PR with these features.
May be latter we can refactor to remove mixins.

Its is better to do incrementally to avoid problems for other users

Josh-ES:
Understood, I'll get to work.

PR #469, opened 2016-09-09
Fixed #465.

### LINKED ISSUE #465: slide width bug when upgrading to 0.14
I upgraded to 0.14 on my development environment today and my slides started getting very large width values and not showing multiple slides at once.

my settings:
```
settings = {
      dots: false,
      infinite: false,
      speed: 500,
      slidesToShow: 6,
      slidesToScroll: 6,
      waitForAnimate: false,
      // Resolutions
      responsive: [{
        breakpoint: 1510,
        settings: {
          slidesToShow: 6,
          slidesToScroll: 6
        }
      }, {
        breakpoint: 1025,
        settings: {
          slidesToShow: 4,
          slidesToScroll: 4
        }
      }, {
        breakpoint: 610,
        settings: {
          slidesToShow: 3,
          slidesToScroll: 3
        }
      }, {
        breakpoint: 480,
        settings: {
          slidesToShow: 2,
          slidesToScroll: 2
        }
      }, {
        breakpoint: 321,
        settings: {
          slidesToShow: 2,
          slidesToScroll: 2
        }
      }]
    };
```

Josh-ES:
Fixed issue #465, which originates in the helpers file in the update method specifically. For whatever reason the bug is not highlighted by calls to the initialize helper.

 - slideWidth used trackWidth directly in calculations, but this.track is not the same thing as this.

### DIFF
--- a/src/mixins/helpers.js
+++ b/src/mixins/helpers.js
@@ -15,9 +15,9 @@ var helpers = {
     var slideWidth;
 
     if (!props.vertical) {
-      slideWidth = trackWidth/props.slidesToShow;
+      slideWidth = this.getWidth(ReactDOM.findDOMNode(this))/props.slidesToShow;
     } else {
-      slideWidth = trackWidth;
+      slideWidth = this.getWidth(ReactDOM.findDOMNode(this));
     }
 
     const slideHeight = this.getHeight(slickList.querySelector('[data-index="0"]'));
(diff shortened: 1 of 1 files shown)


src/mixins/helpers.js
```diff
@@ -15,9 +15,9 @@ var helpers = {
     var slideWidth;
 
     if (!props.vertical) {
-      slideWidth = trackWidth/props.slidesToShow;
+      slideWidth = this.getWidth(ReactDOM.findDOMNode(this))/props.slidesToShow;
```
akiran: @Josh-ES I think, it should be ReactDOM.findDOMNode(this.track) instead of ReactDOM.findDOMNode(this)

Josh-ES: @akiran I changed this line when I wrote the vertical feature to be ReactDOM.findDOMNode(this.track) - which is stored in the trackWidth variable in the diff on the left - because I had the same thought. That's what seems to have causes this bug in the first place!

akiran: @Josh-ES Slides are missing when we resize the window.
I am investigating the issue

PR #474, opened 2016-09-13
custom paging

akiran:


examples/CustomPaging.js
```diff
@@ -0,0 +1,31 @@
+import React, { Component } from 'react'
+import Slider from '../src/slider'
+import {baseUrl} from './config'
+
+export default class CenterMode extends Component {
+  render() {
+    const settings = {
+      customPaging: function(onClick, i, element) {
```
aecorredor: This should be added to the documentation, as it only shows that `customPaging` only takes an `i` argument.

Issue #475, opened 2016-09-14
How to commit code to this repo?


daiduongbian44:
Hi @akiran 
Have a good day, now i wanna update some code to solve #470 , can you help me? In the
`helpers.js` i need to update as the following:
`autoPlay: function (isInitManually) {
    if (this.state.autoPlayTimer) {
      return;
    }
    var play = () => {
      if (this.state.mounted) {
        var nextIndex = this.props.rtl ?
        this.state.currentSlide - this.props.slidesToScroll:
        this.state.currentSlide + this.props.slidesToScroll;
        this.slideHandler(nextIndex);
      }
    };
    if (isInitManually == true || this.props.autoplay) {
      this.setState({
        autoPlayTimer: setInterval(play, this.props.autoplaySpeed)
      });
    }
  },`

`componentWillReceiveProps: function(nextProps) {
    if(nextProps.autoplay == true) {
      this.autoPlay(true);
    } else {
      this.pause()
    }
  },`
Hope you help me.
Thank you very much!

Issue #470, opened 2016-09-12
Update autoplay option, slider does not re-render :(


daiduongbian44:
Hi all, hi @akiran,
First, thanks very much for React-slick :D
Now, i'm doing with this one, and it has an issue, when Autoplay option is changed from True to False and then to True, it does not work although the props of Slider has been changed too.
I don't know what i'm doing wrong, hope you suggest for me :D
Thanks.

TylorShin:
+1

akiran:
Fixed this in f68bb3e68a6997aa018fa60c483be06100ec2a98

saurishkar:
Still facing the issue.

OS : Mac OSX High Sierra
Browser: Chrome
react-slick version: 0.23.1

Issue: Setting autoplay variable from `true` -> `false` -> `true`, the carousel does not autoplay.
Checking the state params fo the inner slider component, the **`autoplaying`** param shows **paused** instead of **playing**

![image](https://user-images.githubusercontent.com/19742306/43944775-b1ffdff4-9c9d-11e8-8419-b1aee45d9bcd.png)

![image](https://user-images.githubusercontent.com/19742306/43944786-bbea5760-9c9d-11e8-81bf-be04a35d937d.png)

PR #476, opened 2016-09-14
Issue #470 #475

daiduongbian44:


src/mixins/helpers.js
```diff
@@ -302,7 +302,7 @@ var helpers = {
         this.slideHandler(nextIndex);
       }
     };
-    if (this.props.autoplay) {
+    if (isInitManually == true || this.props.autoplay) {
```
blairanderson: should clearInterval before setting a new one to make sure that all callbacks are accounted for.

akiran:
This is handled in #526

PR #493, opened 2016-09-27
use loose setting in es2015  in .babelrc

gemscng:


akiran:
Couldn't merge this because of package name. 

But add you change

PR #494, opened 2016-09-27
Custom unslick

blairanderson:
This PR adds unslick customizations. 

- allows for a custom `Unslick` component to be passed in as a prop and used when the slider is unslicked. 
- allows for a responsiveUnslick function to be defined for each breakpoint. If the callback is defined it is used to determine if the slider should switch to unslick

### DIFF
--- a/examples/Responsive.js
+++ b/examples/Responsive.js
@@ -1,6 +1,17 @@
 import React, { Component } from 'react'
 import Slider from '../src/slider'
 
+// return a custom unslicked carousel
+class UnSlick extends Component {
+  render(){
+    return (
+      <div className="my-unslick">
+        {this.props.children}
+      </div>
+    );
+  }
+}
+
 export default class Responsive extends Component {
   render() {
     var settings = {
@@ -10,6 +21,7 @@ export default class Responsive extends Component {
       slidesToShow: 4,
       slidesToScroll: 4,
       initialSlide: 0,
+      unslick: UnSlick,
       responsive: [{
         breakpoint: 1024,
         settings: {
--- a/examples/Responsive.js
+++ b/examples/Responsive.js
@@ -32,10 +32,13 @@ export default class Responsive extends Component {
         }
       }, {
         breakpoint: 600,
+        responsiveUnslick: function({children,newProps,props,state}){
+          return children.length < newProps.settings.slidesToShow;
+        },
         settings: {
           slidesToShow: 2,
-          slidesToScroll: 2,
-          initialSlide: 2
+          slidesToScroll: 1,
+          initialSlide: 2,
         }
       }, {
         breakpoint: 480,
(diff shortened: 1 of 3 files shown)


examples/Responsive.js
```diff
@@ -10,6 +21,7 @@ export default class Responsive extends Component {
       slidesToShow: 4,
       slidesToScroll: 4,
       initialSlide: 0,
+      unslick: UnSlick,
```
blairanderson: Custom unslick component

examples/Responsive.js
```diff
@@ -20,10 +32,13 @@ export default class Responsive extends Component {
         }
       }, {
         breakpoint: 600,
+        responsiveUnslick: function({children,newProps,props,state}){
```
blairanderson: new breakpoint responsive unslick callback

examples/Responsive.js
akiran: Do we really need this feature added to the react-slick.
It can be easily implemented by unmounting Slider and mounting UnSlick component

blairanderson: @akiran I think its pretty necessary.  `react-slick` decides when to unslick at runtime based on responsive widths and its nice to not duplicate that logic both outside/inside the slider.

src/slider.jsx
```diff
@@ -76,11 +71,30 @@ var Slider = React.createClass({
       return !!child
     })
 
+    if (this.state.breakpoint) {
+      newProps = this.props.responsive.filter(resp => resp.breakpoint === this.state.breakpoint)[0];
+      // allow for custom unslick function instead of strictly static settings
+      unslick = newProps.settings === 'unslick' || (
+        typeof newProps.responsiveUnslick === 'function' &&
+          newProps.responsiveUnslick({children,newProps,state: this.state, props: this.props})
```
blairanderson: this is the 2nd feature. allowing a function to be passed in from settings which decides when to unslick.

examples/Responsive.js
akiran: @blairanderson  Can you give me an use where this feature feature will be useful ?

If you want to show the slide vertically on the smaller devices, we already have vertical feature. Check this example
https://github.com/akiran/react-slick/blob/master/examples/VerticalMode.js
You can set vertical in responsive settings.

Let me know if this helps

blairanderson: It doesn't. 

example:
- I have N number of slides on every given page load. randomly from 2 to 6. 
- slides are not variable width, but they have a min-width and max-width which is declared by responsive settings. 
- On a wide screen the carousel settings `slidesToShow: 6` down to small `slidesToShow: 2`

the problem is: If `slide.count < settings.slidesToShow` then the carousel looks pretty silly: https://jsfiddle.net/20bumb4g/1388/

If i'm on a wide screen and 2 are returned from the server, the best way i've found to handle is converting to `UnSlick` and managing widths inside my Unslick component.  

I've built it into react-slick because it already contains all the necessary responsive utilities and breakpoint logic. Seems really silly to duplicate that inside & outside my component.

akiran:
This seems to be specific to your usecase. 
I believe this should be handled with logic outside react-slick library.

I will re-visit if more people request for this usecase

Issue #390, opened 2016-08-03
Lazy load an offset amount of images


gkilian:
### Proposal:

Allow for an offset of images to be lazy loaded, rather than just the one.
### Problem:

Currently when swiping through a set of images with `lazyLoad` set to `true`,  the next image in the queue doesn't start loading until you've committed to the swipe. This causes a second (like in the react-slick examples) or more (when using large images) of no image loaded in the slide.
### Solution:

By implementing something along the lines of `lazyLoadOffset`, we can add a ranges of images to be loaded from the start. Then progressively load the others as the user continues to swipe.

I'm not sure if the community is interested in this functionality. My team has implemented this in our fork of react-slick and are willing to help port over the code if wanted.

akiran:
This issue is fixed in latest release

gkilian:
@akiran - Can we re-open this? I'm not sure that it's actually resolved. 

Although it does seem to quickly load the next image, it's still not loading the image in advance. Thus, while you are in swiping the next images is not there.

This works fine in the Demo example because the next image is off screen, but it doesn't work so well when images are back to back.

janicklas-ralph:
Any update on this? Would be really nice to have this.

i8ramin:
Hello. Wondering if there was any progress made on this issue

emilbruckner:
This is essential

markoshust:
@gkilian do you have the code for this? appears this wasn't pulled into this repo yet. i could really use this functionality. cheers -m

gkilian:
@markoshust - I do in our Condenast fork. I'll submit a PR and tag the people in the PR for review

SBRK:
Would love this as well

gkilian:
@akiran - Are implementation on Conde requires the use of a added prop `lazyLoadOffset`. A secondary PR of mine was closed due to the fact that it supported a non `slick-js` prop. 

Would this be the case for this issue as well? I would hate to write a PR if the end result is the same.

laveesingh:
I'm afraid we cannot accept props that deviate from slick libary as for now.

gkilian:
@akiran - can we have a discussion on how we can work together to rectify these issues, instead of denying them and closing. Our end goal at Conde is to move back to this library over mantaining our fork. In doing so, we need to rectify these issues. 

From the conversations on this issue and in https://github.com/akiran/react-slick/issues/429 - it seems that this plagues more than just our organization. 

### Alternative solution
[...]

PR #495, opened 2016-09-28
good lazyLoaded

Asgator:
UX lazyLoad!

@akiran I add props `preLoad`  he is dont afectid in default logic lazyLoad but  adds new features.  It allows you to pre-load the next component


ddhhz:
This should fix #390.

PR #498, opened 2016-09-30
fixed flickering on window  resize with responsive and autoplay

stanislav-ermakov-roi:
There was a lot of flickering when using responsive option and changing size with autoplay enabled.
This fix eliminates the problem.


akiran:
@stanislav-ermakov-roi Thanks for PR

PR #503, opened 2016-10-02
fixed centerMode strange padding and added SidePadding feature for convenience

renddiyeh:


akiran:
@renddiyeh we already have centerPadding.  why do we need sidePadding option ?

akiran:
@renddiyeh We are trying to follow, API of slick

Since sidePadding option is not present in slick, we can't add it

renddiyeh:
OK, i understand that slick don't need SidePadding.
For the CenterMode, i'm trying to fix is the miscalculation of width, as you can see from current example, CenterMode isn't centerd.
image

And my fix is like this
image

This is more like CenterMode supposed to be.
I can send a PR just for CenterMode correction.

akiran:
Yes please send a PR for that fix

PR #507, opened 2016-10-04
Make CenterMode centering

renddiyeh:


SerbanC:
Hi, is this the PR from https://github.com/akiran/react-slick/pull/503? I've been struggling for half a day to fix this bug, it would be wonderful if this gets fixed

akiran:
@renddiyeh Thanks for PR.
Will make release soon

Issue #508, opened 2016-10-05
Slider container should not have `slick-initialized` class when rendered on server


jonjaques:
AFAIK, the normal slick implementation does not add `slick-initialized` until the JS has run; I believe this class should be applied via `componentDidMount` to avoid FOUC.

gkilian:
Similar to this issue https://github.com/akiran/react-slick/issues/429

laveesingh:
Referenced commit should take care of it.

Issue #273, opened 2016-04-02
Does this work with server side rendering? no 'window' or 'document' dependencies? 


Ajar-Ajar:
I've been trying out too many carousel react components that assume they are running on the client and have 'window' and 'document' dependencies all over the place. This breaks the code trying to integrate it on an isomorphic app...

jonathanewerner:
Nope, it doesn't, also depends on mounted DOM for measurement.

MitchWeiss:
@Ajar-Ajar did you ever find a solution for isomorphic rendering?

Ajar-Ajar:
Isomorphic / universal React web-apps are becoming the standard
To avoid a `ReferenceError: window is not defined`
I'd suggest to wrap all dependencies to window with

```
var canUseDOM = !!(
  typeof window !== 'undefined' &&
  window.document &&
  window.document.createElement
)
if(canUseDOM ){
     // use window dependency...
}else{
  //have an alternative that wouldn't break functionality...
}
```

nicholasdickey:
Really important issue, thanks!

akiran:
This should be fixed in react-slick@0.13.6

Ganasist:
I'm having the same problem with a universal setup. The carousel mostly works when I use the stylesheet tags in <head>, but on first load (server-side render) there are visual bugs that go away by clicking through links (client-side render).

When I try with the npm package instead of stylesheet tags it doesn't work at all.

ypyang237:
Ran into this issue as well. Current solution is as @Ganasist  mentioned, if working with a universal setup, it will only work properly if you include both the stylesheet and the npm package.

alexanthony813:
@ypyang237 I'm having this issue as well. When you say "include both the stylesheet and the npm package", do you mean you are including the CDN for `slick-carousel`, while using npm for `react-slick`? Thanks in advance!

ypyang237:
@alexanthony813  
Yes, you understood correctly. I included the two CDN stylesheets for slick-carousel and had react-slick installed with npm as well to get it working. Hope this helps you!

coreyleelarson:
I have react-slick working with server-side rendering, however I get the following error when using certain responsive settings. I believe it has to do with features only available in my "responsive" breakpoint (i.e. dots):

**Settings:**
```javascript
const settings = {
  dots: false,
  draggable: false,
  fade: true,
  infinite: true,
  swipe: false,
  responsive: [
    {
      breakpoint: 767,
      settings: {
[...]

PR #509, opened 2016-10-05
Server-Rendering new aproach for default sliders

joshrezende:
So i see many people who needs the component running on server side(include myself), I made some changes in the lifecycle and main logic, now on default slides we use percentage based size, this way we don't need get parent size for calculate childs width. 

the others types continue to work, running normal flow of events.

some issues about this: #273 #508


lucasvmiguel:
Good Idea!

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-slick",
-  "version": "0.14.5",
+  "version": "0.14.5-beta.0",
   "description": " React port of slick carousel",
   "main": "./lib",
```
afgoulart: Add key:
"files": [ "lib/" ],

for send only available files.

candreoliveira:
+1

@akiran how long to merge the PR?

frankleng:
👍

imthehell:
is the PR been merged?

joshrezende:
@imthehell not yet, you can test from my repo, there some new improves I made on another parts, but need to push.

imthehell:
Ok, many thanks :-) Looking forward for the merge

bforma:
Any updates here? I'm eager on rendering react-slick universally..

joshrezende:
@bforma sadly,  not yet... 
I'm wondering in make merge of another's PR here on Main repo and make my fork more active with the users who are making changes, what you think?

imthehell:
I think is a good idea.

I have found another issue, the "centerMode: true" option is not working well when rendered server side. Looking on the code, the problem should be here

`ar getSlideStyle = function (spec) {
  var style = {};
  var percentUnit = spec.percentUnit;
  var sizeUnit = percentUnit ? '%' : 'px';

  if (spec.variableWidth === undefined || spec.variableWidth === false) {
    style.width = spec.slideWidth + sizeUnit;
  }

  if (spec.fade) {
    style.position = 'relative';
    style.left = -spec.index * spec.slideWidth;
    style.opacity = (spec.currentSlide === spec.index) ? 1 : 0;
    style.transition = 'opacity ' + spec.speed + 'ms ' + spec.cssEase;
    style.WebkitTransition = 'opacity ' + spec.speed + 'ms ' + spec.cssEase;
  }

  return style;
};`

The width of the slide is calculated with the % on server side and "px"on the client and the calculated widths are different. 
The following function has been added in order to calculate the slide width server-side

`getSlidePercentWidth: function(props) {
    var slideWidth = 100/(props.children.length + 2*props.slidesToShow);

    return slideWidth;
  },`

but it is not working with "centerMode: true"

Anyway, many thanks for the great work!

joshrezende:
@iamthehell, centerMode stay in legacy functions,  I'm going to see this problem, tomorrow and post my concerns, but I think that I place him on legacy because this px units necessary for calculation.

imthehell:
Yes it is placed on legacy functions, when I removed it from legacy I have found this problem of pixel calculation. Don't now if it is possibile to calculate the px server side..

mkreidenweis:
What's the status on this PR? 

I'm thinking about creating a fork to at least do the initialization of `slideCount` on the server side in order to get the dots/pagers rendered on the server. 

If you're no planning to merge this, what would you think about a more minimal approach which just fixes the `slideCount` issue and maybe #508?

joshrezende:
@mkreidenweis the PR still the same, you like I'm wondering too on create a fork, I've made some others updates on the component but using local, do you like to join forces?

We have anothers dev's seeking for server render too.

aperkaz:
Hello, is there any news about this? We really need to have server side rendered carousel with react... 
Thanks

aperkaz:
Hello @laveesingh ,
Why has this pull PR been closed?
Thanks

Issue #413, opened 2016-08-15
The setInterval should be cleared after each next/prev button click


philiptzou:
For example, using the default `autoplaySpeed` which is 3000ms and enabling [`autoplay`](https://github.com/akiran/react-slick/blob/08d65e427b87bbb45e8081076d4c6ba21e1f4103/src/mixins/helpers.js#L253). If one clicked the next arrow button at about 2000ms, `react-slick` will switch to next page in 1000ms but not 3000ms. This problem also happens when you click any of the dot buttons. I tested this on [slick-carousel](http://kenwheeler.github.io/slick/) but it didn't happen.

vishwanatharondekar:
@akiran This needs to be closed since its fixed in https://github.com/akiran/react-slick/pull/526

akiran:
Thanks @vishwanatharondekar 

Closing this issue

PR #512, opened 2016-10-12
restart autoPlay on changeSlide

data-doge:
hi @akiran thx for your work on this module

this PR aims to address https://github.com/akiran/react-slick/issues/413


akiran:
This is addressed in  #526

PR #516, opened 2016-10-18
feat: add support for setting autoplaySpeed in the each slide individually

vishwanatharondekar:
We have a requirement where we need to set different autoplaySpeed to each slide depending on type of the slide.

I am not sure whether you will want to keep this as a part of the main repo. But looks like will be useful for someone having similar requirement. 

I will build upon this pull request with example and documentation if you are interested.

PS : This also fixes one of the problem in autoplay where slide always changes based on interval time set even if user is interacting with the slides. Ideally time should reset if user has changed the slide himself.

### DIFF
--- a/src/mixins/helpers.js
+++ b/src/mixins/helpers.js
@@ -179,7 +179,7 @@ var helpers = {
         this.props.beforeChange(this.state.currentSlide, targetSlide);
       }
 
-      this.autoPlay();
+      this.autoPlay(this.props.children[targetSlide].props.autoplaySpeed || this.props.autoplaySpeed);
       return;
     }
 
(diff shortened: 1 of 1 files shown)


akiran:
We might not support per slide autoplaySpeed at the moment.

akiran:
@vishwanatharondekar Thanks for the PR.

I incorporated your remaining part of the fix in  https://github.com/akiran/react-slick/pull/526

vishwanatharondekar:
No problem. Happy to help. Any particular reason we are not looking at incorporating the feature.

akiran:
We trying to closely follow the behaviour of original slick project.
This is not supported in it

alexanderfrey:
Well, then lets make it better !

PR #529, opened 2016-10-31
Only suppress default if swipe is enabled

jeffh:
Issue: Having items with click handlers inside react-slick do not get
triggered when the option swipe is set to false.

This is because `swipeEnd` called `preventDefault()`.

Tested on: iPad mini 4 & iPad Air 2 Simulator


Shion093:
Hi, how can I access this fix? Is not in my src/mixins/event-handlers.js version 0.14.5

Thanks.

jeffh:
This PR will be in the next release. Alternatively I guess you can try installing from git: https://docs.npmjs.com/cli/install

PR #554, opened 2016-11-22
fixing numerous bugs when calculating the slide offset for finite sets

ryanbagwell:
It looks like my [previous PR](https://github.com/akiran/react-slick/pull/425) didn't address all possible conditions for a finite carousel. This one seems to. I tested it with all different kinds of combinations of slidesToScroll, slidesToShow and the total number of slides, and it seems to work.


ryanbagwell:
@akiran any chance you could take a look at this when you have a moment?

PR #559, opened 2016-11-28
Enable server-side rendering of dots/pagers

mkreidenweis:
The `slideCount` and `currentSlide` values can easily be calculated during server-side rendering already. 
When we do that, it will get the the dots/pagers rendered on server side already in isomorphic apps, thus preventing any "jumping" of the page content after JavaScript has been initialized. 

It would fix part of what #509 tries to do.

### DIFF
--- a/src/inner-slider.js
+++ b/src/inner-slider.js
@@ -50,6 +50,10 @@ export var InnerSlider = createReactClass({
         lazyLoadedList: lazyLoadedList
       });
     }
+
+    if (typeof window === 'undefined') {
+      this.serverInitialize(this.props);
+    }
   },
   componentDidMount: function componentDidMount() {
     // Hack for autoplay -- Inspect Later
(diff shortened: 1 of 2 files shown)


gilligan:
Hey @laveesingh - could you perhaps elaborate on why you decided to close this issue?

Was the problem it was supposed to fix solved elsewhere ? Was there something about the PR that wasn't acceptable?

We are running a [fork of react slick](https://github.com/holidaycheck/react-slick) but we would love to try to get our improvements upstream if they make sense to you to have everyone benefit from it. So some feedback would be very appreciated. Thank you!

laveesingh:
Since the library has advanced quite a bit, we are closing old enough PRs after taking issues into account. All such issues will have effective fixes asap.
@gilligan we would love to have effective changes, please feel free to open a PR.

PR #565, opened 2016-11-30
Fix flashes during sliding between slider head & tail in centerMode

MPjsx:


szimek:
@MikeJSX Hey, thanks for taking care of it. One question though - is the only required change actually the one here https://github.com/akiran/react-slick/pull/565/files#diff-0d92326ea3a7159fcb03bae37efad9b3R108 ? Because you also added the whole `lib` folder and I'm not sure if anything has changed there.

MikeJSX:
@szimek Hi, yes, I see this issue two so I used your code to fix this. It works now as expected, thanks. Those folders are from built version. I use my github repo as package.json dependency.

szimek:
@MikeJSX Today I released a version with my both fixes (i.e. #567 and #568)  - [@freeletics/react-slick](https://www.npmjs.com/package/@freeletics/react-slick), so that I can avoid putting the built version in the repo 😉

DavisChang:
Do Conflicting files have been resolved? Thanks~

ChazUK:
4.5 years on and I'm seeing the flash still, was this ever fixed?

Issue #557, opened 2016-11-23
Not enough slick clone slides causing flashing during sliding between slider head & tail


jackytwh:
Glad if anyone can help, thanks a lot.

```
var settings = {
    centerMode: true,
    centerPadding: '100px'
}
```

https://jsfiddle.net/jackytwh/20bumb4g/1644/

MikeJSX:
I created [PR](https://github.com/akiran/react-slick/pull/565) for this issue. Check if you'd like

jackytwh:
@MikeJSX , thanks for looking into the issue.
However, the infinite value is default to be "YES" though i didn't state it.
Therefore, i wonder if you are fixing the code in the right place.
The issue seems still exist. Kindly take a look sample code in jsfiddle for details.

szimek:
@MikeJSX For me your fix almost works 😉  The number of slides is correct now, but with centerMode + infinity I'm getting invalid offset for the whole track, i.e. the center slide is not marked as center, but the one on the right. Looking at the source code of the original slick carousel it seems one needs to adjust track left offset by additionally subtracting slide width, i.e. change [this](https://github.com/akiran/react-slick/blob/master/src/mixins/trackHelper.js#L118) line to:
```js
slideOffset += spec.slideWidth * Math.floor(spec.slidesToShow / 2) -  spec.slideWidth;
```
Now everything seems to be fine, at least in my particular case.

laveesingh:
I am not able to reproduce this anymore. Maybe it got fixed around 8feb5cd. 
If someone is still able to regenerate it, please feel free to suggest reopening the issue.

PR #568, opened 2016-12-02
Fix slide popping up in infinite + centerMode

szimek:
It fixes the issue mentioned in #557. 

It's based on the original code from slick carousel (and @MikeJSX PR https://github.com/akiran/react-slick/pull/565, but it fixes the issue where adding one more cloned slide, caused all slides to be incorrectly offset by 1).

Here's a demo of the broken version: https://jsfiddle.net/30rnf9kn/ (just move slides a few times in the same direction to see the issue)
Here's a demo of the fixed version: https://jsfiddle.net/094o7tcz/

### DIFF
--- a/src/mixins/trackHelper.js
+++ b/src/mixins/trackHelper.js
@@ -115,7 +115,7 @@ export var getTrackLeft = function (spec) {
 
   if (spec.centerMode) {
     if(spec.infinite) {
-      slideOffset += spec.slideWidth * Math.floor(spec.slidesToShow / 2);
+      slideOffset += spec.slideWidth * Math.floor(spec.slidesToShow / 2) -  spec.slideWidth;
     } else {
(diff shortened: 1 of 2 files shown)


JSZYN:
Looking forward to this fix

szimek:
@JSZYN I released it as @freeletics/react-slick, so if you really need it, you can use my package before this PR is merged to the official repo.

Amazing-Space-Invader:
> commented on 2 Dec 2016
Why u don't merge it?

@szimek can I install your package via NPM? Give me link please

szimek:
@Amazing-Space-Invader https://www.npmjs.com/package/@freeletics/react-slick. Though I haven't updated it in a while, so it might have some other bugs that were already fixed here. But we're still using it in our app.

Issue #547, opened 2016-11-16
Can not click(onClick) button in child component when swipe props is false on mobile.


leekyuchang:
I'm using react-slick 0.14.5 version.
I can't click button in child component when swipe props is false on mobile.
And also, swipeToSlide is not working on mobile.

```javascript
const settings = {
      dots: false,
      arrows: false,
      draggable: false,
      // swipeToSlide: false,
      swipe: false
};

<Slider {...settings}>
  <div>
    <Main />
  </div>
</Slider>
```

mnt:
I'm also experiencing the same issues. Example of the issue illustrated here:
https://jsfiddle.net/20bumb4g/1627/

leekyuchang:
I modify swipeEnd code(delete e.preventDefault();). So I can click button in child component.
```javascript
swipeEnd: function swipeEnd(e) {
    if (!this.state.dragging) {
      // e.preventDefault();
      return;
    }
    ...
```

RyoIkarashi:
Does the modified code @leekyuchang mentioned above have any side effects?
If no, then I'm gonna create a PR for it. So can you, @akiran just merge the PR?

laveesingh:
Similar issues have been fixed in previous releases, please try the latest version.
If it still persists, please create a codesandbox demo for the same and reopen it.

PR #571, opened 2016-12-03
Enable click event even when the swipe option is false

RyoIkarashi:
https://github.com/akiran/react-slick/issues/547

### DIFF
--- a/src/mixins/event-handlers.js
+++ b/src/mixins/event-handlers.js
@@ -41,7 +41,7 @@ var EventHandlers = {
 
     this.slideHandler(targetSlide);
   },
- 
+
   // Accessiblity handler for previous and next
   keyHandler: function (e) {
     //Dont slide if the cursor is inside the form fields and arrow keys are pressed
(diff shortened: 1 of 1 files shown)


martinpesout:
@akiran this PR is also important for use. We have set `swipe` to `false` and cannot click (on touch devices) on elements inside slick slides. This will fix it. Do you think that you can merge it?

Thanks to @RyoIkarashi for this fix.

malliapi:
@akiran I know you must be busy, but is it possible to have a look at seeing if you agree with merging this? Many thanks :-)

malliapi:
ok so to anyone in need of this urgently i have published here https://www.npmjs.com/package/react-slick-2

martinpesout:
Hi, is it possible to merge this PR?

laveesingh:
Similar issues have already been fixed in previous releases, please try the latest version. If the issue still persists for you, please create a codesandbox demo and reopen it.

Issue #488, opened 2016-09-23
Add animate argument to slickGoTo method


shendrick:
So it looks like the slick library accepts a second argument of a boolean to animate or not..

``` javascript
    Slick.prototype.goTo = Slick.prototype.slickGoTo = function(slide, dontAnimate) {

        var _ = this;

        _.changeSlide({
            data: {
                message: 'index',
                index: parseInt(slide)
            }
        }, dontAnimate);

    };
```

But it doesn't seem to be accepted in react-slick...

``` javascript
slickGoTo: function slickGoTo(slide) {
    this.innerSlider.slickGoTo(slide);
},
```

PsiXdev:
Hello. How are things going with this issue? I also need to use this functionality.

edencorbin:
This would also be huge for me, any progress?

goncalomarques:
Any updates on this?

mricketts:
Checking in here, is there no current way to have the slickGoTo method animate?

peterph-am:
Any updates on this issue? I would like to animate the slickGoTo method too.

laveesingh:
The referenced commit should take care of it.

leomyth330:
Could you release new version with this update?

PR #579, opened 2016-12-15
Add dontAnimate argument to slickGoTo method

louisptremblay:
Fixes issue #488
Also fixes the example slickGoTo which wasn't working because it was passing a string instead of a number.


edencorbin:
Can we get this added to merged, we need this feature!

bratva:
Please, do it

Leapfrognz:
+1 for this. original slick has it, this port needs it too.

aperkaz:
Hi,
Is there any reason why this hasnt been merged yet? 
I do not understand... :slightly_frowning_face:

laveesingh:
Can't support this feature. 
Same functionality can be achieved by managing an additional state in wrapper component.

PR #581, opened 2016-12-17
Adds fallback 0 to getWidth and getHeight in case of undefined.

coreyleelarson:
We are using react-slick for an image carousel and our mocha tests are throwing random NaN console.logs:
```
Warning: `NaN` is an invalid value for the `width` css style property. Check the render method of `ProductDetailImage`.
```

Our tests are not currently failing, but we would like this message to disappear.

We've narrowed it down to the **getWidth** and **getHeight** functions returning undefined because our test images are mocked.

**elem.getBoundingClientRect().width** and **elem.getBoundingClientRect().height** are returning 0, but falling back to **elem.offsetWidth** and **elem.offsetHeight** which are returning undefined. We felt this should fallback to 0 in that case.

### DIFF
--- a/src/mixins/helpers.js
+++ b/src/mixins/helpers.js
@@ -94,10 +94,10 @@ var helpers = {
     });
   },
   getWidth: function getWidth(elem) {
-    return elem.getBoundingClientRect().width || elem.offsetWidth;
+    return elem.getBoundingClientRect().width || elem.offsetWidth || 0;
   },
   getHeight(elem) {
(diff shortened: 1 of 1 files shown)


jblok:
I am getting the same error message in my mocha test logs. Thanks for the PR @coreyleelarson, this fixed it :)

@akiran merge to master would be awesome

coreyleelarson:
Can a new version be published to npm so we may pull in this change?

arjunskumar47:
What if `elem` is `null` or `undefined` ?

PR #592, opened 2017-01-02
Add beforeSlick, beforeUnslick callbacks

hinok:
Hey,
I'm building small wrapper around `react-slick` and I've needed a callback when unslick is done.
I added `beforeUnslick` and `beforeSlick`. Feel free to propose different names for callbacks.
I was trying to pick something that will be fit into the current naming like: `beforeChange`. `afterChange`.

I fixed also small difference when `unslick` is set and how `settings` are created.

### DIFF
--- a/src/slider.js
+++ b/src/slider.js
@@ -7,6 +7,8 @@ import json2mq from 'json2mq';
 import ResponsiveMixin from 'react-responsive-mixin';
 import defaultProps from './default-props';
 
+var isFunction = f => typeof f === 'function';
+
 var Slider = React.createClass({
   mixins: [ResponsiveMixin],
   innerSlider: null,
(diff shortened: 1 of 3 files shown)


src/slider.js
```diff
@@ -61,7 +63,7 @@ var Slider = React.createClass({
     var newProps;
     if (this.state.breakpoint) {
       newProps = this.props.responsive.filter(resp => resp.breakpoint === this.state.breakpoint);
-      settings = newProps[0].settings === 'unslick' ? 'unslick' : assign({}, this.props, newProps[0].settings);
+      settings = newProps[0].settings === 'unslick' ? 'unslick' : assign({}, defaultProps, this.props, newProps[0].settings);
```
hinok: Here `defaultProps` has never been used but if you look at non-responsive version, they are used:

```javascript
settings = assign({}, defaultProps, this.props);
```

akiran:
I  think, It is better to not add these methods in this library

This functionality can be achieved with componentWillMount/constructor and componentWillUnMount on component component wrapping react-slick component

hinok:
@akiran Ahh I didn't think about it, you are right ;)

PR #602, opened 2017-01-16
Pass currentSlide and slideCount props to arrow components

psirenny:
My use case for this was to change the last slide's arrow from saying "Next" to "Done". The default arrows have access to these props and use them to set a disabled class. It seems only fair that custom arrows have these props as well 😄 .

I found it difficult to use the `beforeChange` handler alone to achieve this without causing the entire carousel to re-render. That was bad because it prevented the slide transitions from animating.

### DIFF
--- a/src/arrows.js
+++ b/src/arrows.js
@@ -23,6 +23,8 @@ export var PrevArrow = React.createClass({
       key: '0',
       'data-role': 'none',
       className: classnames(prevClasses),
+      currentSlide: this.props.currentSlide,
+      slideCount: this.props.slideCount,
       style: {display: 'block'},
       onClick: prevHandler
     };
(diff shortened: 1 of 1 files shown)


petrbela:
This is now causing in 0.14.6:

```
warning.js?8a56:36 Warning: Unknown props `currentSlide`, `slideCount` on <button> tag. Remove these props from the element. For details, see https://fb.me/react-unknown-prop
    in button (created by PrevArrow)
    in PrevArrow (created by InnerSlider)
    in div (created by InnerSlider)
    in InnerSlider (created by Slider)
    in Slider (created by Categories)
    in div
    in Categories (created by Unknown)
    in MediaQuery
    in div
    in Unknown
    in div
    in div
    in DayPlanner
    in div
    in div
    in Planner (created by PlannerHandler)
    in PlannerHandler
```

L3V147H4N:
same warning here

PR #605, opened 2017-01-20
Adds slickPause() and slickPlay() methods

xrs1133:
@jeffreznik Did all the work. I just removed some of the props.
I bumped the version as well


vamshi29292:
I'm using this fork and it has an issue with pauseOnHover. I'm using it for a carousel of videos which needs to pause on playing a video. I have pauseOnHover set to true. Whenever I play a video, the carousel pauses and as soon as I hover onto it, it starts again.

Issue #534, opened 2016-11-04
TypeError: undefined is not an object (evaluating 'targetSlide.offsetWidth')


jcady:
https://jsfiddle.net/20bumb4g/1542/

Drag slider to the right past bounds to see JS error.

akiran:
@jcady  You are seeing this issue only with centerMode = true and infinite = false right ?

jcady:
@akiran Correct. Also seems to disappear if I set variableWidth = false.

roman-yakobnyuk:
This happens for me too. Is there a fix for it yet?

laveesingh:
Similar issues have been fixed in previous releases, please try the latest version. If the issue still persists for you, please create a codesandbox demo and reopen it.

PR #609, opened 2017-01-23
Handle edge cases and catch errors

roman-yakobnyuk:
This patch prevents slider from trying to scroll out of boundaries and cause offset errors. This behaviour happened with following settings: 

```
let settings = {
            infinite: false,
            slidesToScroll: 1,
            swipeToSlide: false, // Also with true
            centerMode: true
        };
```

This patch allows slider to snap back to previous slide if it run out of slides. 
Also added correct error handling for cases when 'swipeToSlide' enabled, however it still does not work too well and needs further refactoring. 

This patch should solve issue #534 and #357 and maybe other related issues too.

Please merge and publish as soon as possible, we are using it in production. 

Thank you!

### DIFF
--- a/src/mixins/event-handlers.js
+++ b/src/mixins/event-handlers.js
@@ -226,7 +226,13 @@ var EventHandlers = {
         return true;
       });
 
-      const slidesTraversed = Math.abs(swipedSlide.dataset.index - this.state.currentSlide) || 1;
+      var slidesTraversed;
+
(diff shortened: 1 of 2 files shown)


laveesingh:
Similar issues have been fixed in previous releases, please try the latest version. If the issue still persists for you, please create a codesandbox demo and reopen it.

PR #638, opened 2017-02-16
Remove `parseInt` from `changeSlide` index method

ericfreese:
`slidesToScroll` option allows non-integer (fractional) values.

Why restrict `slickGoTo` to only integers?

Edit: Just occurred to me, some folks may be passing strings into `slickGoTo`. In that case, how about using `parseFloat` instead of `parseInt`?


akiran:
Used  Number instead of parseInt/parseFloat

PR #641, opened 2017-02-17
Fix getHeight and getWidth erroring when no slides exist

samuelmaddock:
The `getHeight` and `getWidth` helpers will throw an error if no slides are present. The element passed into these helpers is the result of [a querySelector](https://github.com/akiran/react-slick/blob/8c3652b727bccb1ad75bf376cdcaaf93100b7b5b/src/mixins/helpers.js#L24) with no result (`null`).

### DIFF
--- a/src/mixins/helpers.js
+++ b/src/mixins/helpers.js
@@ -94,10 +94,10 @@ var helpers = {
     });
   },
   getWidth: function getWidth(elem) {
-    return elem.getBoundingClientRect().width || elem.offsetWidth || 0;
+    return elem && (elem.getBoundingClientRect().width || elem.offsetWidth) || 0;
   },
   getHeight(elem) {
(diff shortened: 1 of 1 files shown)


raviteja83:
Please  merge this pull request @akiran

akiran:
Fixed it in master

pcooney10:
Hey @akiran, thank you for fixing this. When you get a chance, can you bump the version and push it to npm? Thanks !

akiran:
I will make a release next Sunday with fixes for few other issues

williamboman:
@akiran Could you /cc us here when you've released new version? Thanks 🎉!

williamslihf:
the problem still exist, becasue the slickList is null, so the slickList.querySelector('[data-index="0"]') will throw TypeError.

laveesingh:
This could have been caused by uncleared timeout callbacks, but this has been fixed and the changes will be released soon.

PR #646, opened 2017-02-20
Fixed animation timeout IDs never getting set/cleared

samuelmaddock:
The animation changes in #125 forgot to include the arrow function changes. Thus, the callback ID never gets set on the correct context. This can lead to `setState` getting called after the component has been unmounted.


laveesingh:
@samuelmaddock since we're using slideHandler as a mixin, I don't think we'd have that context problem as you mentioned. But it's not a bad idea to use arrow functions instead, so I'll merge the PR.

Thanks for your efforts.

PR #661, opened 2017-03-03
fix centerMode issue

Apkawa:
fix calculation issue in initial stage for short width, as example 
```
<div style={{witdh: 500}}><VariableWidth /></div>
```

and fix calculate centerMode issue


laveesingh:
Most of the centerMode issues have been fixed. This PR contains dist and lib changes, hence can't be merged right in.

Issue #623, opened 2017-02-06
Warning: Unknown props `currentSlide`, `slideCount` on <div> tag


maslianok:
version 0.14.6

Error
```
Warning: Unknown props `currentSlide`, `slideCount` on <div> tag. Remove these props from the element. For details, see https://fb.me/react-unknown-prop
    in div (created by Carousel)
    in PrevArrow (created by InnerSlider)
    in div (created by InnerSlider)
    in InnerSlider (created by Slider)
    in Slider (created by Carousel)
```

The bug was introduced in this commit https://github.com/akiran/react-slick/commit/436e3052855613a49478382b51743f8bbef9bf15

You added 2 custom props. But 10 lines below you [use this object as props for the `div` element](https://github.com/akiran/react-slick/blob/436e3052855613a49478382b51743f8bbef9bf15/src/arrows.js#L36)

aiduryagin:
@maslianok, add "transform-object-rest-spread" plugin to .babelrc file

intpp:
@aiduryagin it doesn't help

aiduryagin:
@intpp plugin "transform-object-rest-spread" for transpilation his source code into dist and lib folders :)

dennyferra:
Ran into this issue as well - https://github.com/maslianok/react-slick/commit/8ed285ee0ec97f1de7fdad36610dfb5fe764da70 fixes it, will there be a pull request?

wellnine:
It doesn't help.  I also have the same issue 
 **Unknown props `currentSlide`, `slideCount` on <button> tag.**   and I added this plugin but  it didn't fix anything, unfortunately.

ystarlongzi:
[transform-object-rest-spread](https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-object-rest-spread)
> This plugin allows Babel to transform rest properties for object destructuring assignment and spread properties for object literals.

According to the description, the plugin is not intended to solve `react unkown props` problem.

maslianok:
I've updated the PR.
`rest` operator is replaced with `Object.assign`.

gilbarbara:
I don't think the problem is with spread at all..
HTML Elements shouldn't have these custom props. From the [docs](https://facebook.github.io/react/warnings/unknown-prop.html):
> You are using a non-standard DOM attribute on a native DOM node, perhaps to represent custom data. If you are trying to attach custom data to a standard DOM element, consider using a custom data attribute as described on MDN.

[...]

PR #662, opened 2017-03-06
Hotfix/arrow props cleanup

NullDivision:
I've cleaned up the arrow elements and left the custom props in for custom arrows.

I didn't see any use for them on the standard buttons.

This should fix issue #623.


NullDivision:
@akiran I've updated Jest and created a couple of tests. I would like to help in fixing the issues you have with the library.

PR #679, opened 2017-03-19
issue:344 - set height once images are loaded

gregnb:
vertical slider would never load correctly because height was not determined in initialize state due to images still being loaded. once images are loaded, just update() state

### DIFF
--- a/src/mixins/helpers.js
+++ b/src/mixins/helpers.js
@@ -20,7 +20,19 @@ var helpers = {
     } else {
       slideWidth = this.getWidth(ReactDOM.findDOMNode(this));
     }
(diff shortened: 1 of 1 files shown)


PhilipBrady13:
any chance we can get this merged? been having same issue using carousel and would like to continue using your tool

Mr-Taco:
wouldnt count on it. can fork repo and use this update though

laveesingh:
I tried myself and found a very related issue, but the solution suggested here doesn't fix the slideHeight issue. If you believe otherwise, please feel free to replicate the issue and request reopen.

Issue #94, opened 2015-08-04
Text blurry in chrome when using "tranform:translate3d(x,y,z)"


HuiguangChu:
could you change it to translateX for WebkitTransform like you did for msTransform?

vramana:
Shouldn't this be a bug in chrome? I am not sure how changing to translateX will solve your problem because they are literally same API except `translate3d` moves in two other directions. Have you tried this out? 

Which version of chrome are you using? Can you show a screenshots of using `translate3d` and `translateX`? Or just simply where you are having the blurry text? Why do you think its caused by `translate3d`?

HuiguangChu:
okay, you can compare two screenshots, one is using translateX and one is
using translate3d.
 Actually, translate3d cause text blurry in chrome is the common issue.


2015-08-04 15:56 GMT+02:00 Ramana Venkata <notifications@github.com>:

vramana:
Can you check if this works http://stackoverflow.com/a/7477166/1117168 ??

HuiguangChu:
thank you, i did that, it did not work. I don't get the point that why you are using 3d, as I can see you set  the y and z is 0. why you don't using translateX?, then you don't need to fix bug for IE as well

vramana:
I am slightly reluctant to do that because I want to know what exactly is the cause of problem and what is the best way to solve it. I am trying to read whatever I can find.

[This question](http://stackoverflow.com/questions/27385126/chrome-font-appears-blurry) states using `translateX` causes problems. According to [this reddit comment](https://www.reddit.com/r/web_design/comments/2z9ub7/css3_translatey_in_webkit_browsers_makes_text/cpgze9v), any type of transform can cause the blur. Sometimes the fixes work, sometimes they don't.

vramana:
@akiran How do you want to solve this?

akiran:
@HuiguangChu Original slick from which we ported this also uses translate3D 
https://github.com/kenwheeler/slick/blob/master/slick/slick.js#L1696

gtqnchev:
I'm experiencing blurry text too. 
It seems to happen when non-integer values are given to translate3d.
Here is an example: [blurry](https://www.dropbox.com/s/yygbhe6n5kbr43t/Screenshot%202016-08-17%2016.14.05.png?dl=0),  [ok](https://www.dropbox.com/s/mcmo3tg376cu772/Screenshot%202016-08-17%2016.14.52.png?dl=0)

[...]

PR #680, opened 2017-03-20
Fix Slide width calculation error

### LINKED ISSUE #221: Slide width calculation error
Hi,

The slideWidth value is wrong (at least not coherent with the original slick library), it should be Math.ceil(trackWidth/props.slidesToShow) to give the same result as slick.

ghostd:
Fix #221 and could also fix (or try to) #94


laveesingh:
The changes have been implemented in commit: e10e4fb

Thanks for the PR though.

Issue #603, opened 2017-01-17
Fade effect doesn't work properly 


decoder10:
When you drag the image it doesn't work.

if you remove this style 
`transform: translate3d(0px, 0px, 0px);`
from  `.slick-track` it will work :)

laveesingh:
Problems related to fade+swipe were solved in the previous release. I'm closing this for now. If someone still faces any issue similar to this, please feel free to open another one.

Issue #585, opened 2016-12-22
Image carousel caught mid-swipe (mobile) or mid-drag (desktop)


rainesinternationaldev:
I understand that this issue has been brought up in #162 and #412, but I'm not sure if it's been fully addressed. Images simply stop scrolling.

#### Initial page load:
<img width="608" alt="screen shot 2016-12-22 at 3 11 10 pm" src="https://cloud.githubusercontent.com/assets/21286106/21438955/82781130-c859-11e6-87c7-a9fa621a9b22.png">

####  Post-swipe:
<img width="606" alt="screen shot 2016-12-22 at 3 16 42 pm" src="https://cloud.githubusercontent.com/assets/21286106/21439001/ae194304-c859-11e6-9045-8eaea952b634.png">

####  Images continue to auto-fade in that position:
<img width="603" alt="screen shot 2016-12-22 at 3 16 45 pm" src="https://cloud.githubusercontent.com/assets/21286106/21439011/b6015bc4-c859-11e6-8b18-b1256e2162da.png">

Settings:
```
{
      dots: true,
      infinite: true,
      fade: true,
      speed: 1000,
      slidesToShow: 1,
      slidesToScroll: 1,
      autoplay: true,
      autoplaySpeed: 5000,
      arrows: false,
      draggable: true,
      swipe: true,
      swipeToSlide: true
};
```
Version: `v0.14.5`

coreyleelarson:
We are having the same issue. We suspect it has something to do with "fade" and "swipe" both being enabled, but haven't come up with a solution.

SrSandeepKumar:
Facing similar issue once i zoom in on image and scroll.

cgalbiati:
Also having the same issue and pretty sure it is caused by fade and swipe together.

laveesingh:
Problems related to fade+swipe were solved in the previous release. I'm closing this for now. If someone still faces any issue similar to this, please feel free to open another one.

PR #686, opened 2017-03-23
Fix for Bug in Slide Position while Using Fade and Swipe

cgalbiati:
Currently, the library does not support the use of fade animation while swipe is enabled.  Using both of these results in slide positioning bugs (as documented https://github.com/akiran/react-slick/issues/585 and https://github.com/akiran/react-slick/issues/603).

I have fixed the bug by only applying the transforms if fade is not active (slide should move across the screen while swiping).

### DIFF
--- a/src/mixins/trackHelper.js
+++ b/src/mixins/trackHelper.js
@@ -30,27 +30,27 @@ export var getTrackCSS = function(spec) {
   }
 
     var style = {
-        opacity: 1,
+      opacity: 1,
     };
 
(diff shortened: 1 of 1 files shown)


ashkassw:
It would be great if this can be merge into the next release.

laveesingh:
Problems related to fade+swipe were solved in the previous release. I'm closing this for now. If someone still faces any issue similar to this, please feel free to open another one.

PR #687, opened 2017-03-27
fix edge drag case

ayal:
this prevents edgedrag when scrolling vertically but slightly right/left
swipeDirection is "vertical" because of angle (see swipeDirection) so above edgeDrag protection doesn't work 
in this case just leave swipeLeft at same position


laveesingh:
could you generate the issue that you're trying to fix, and please feel free to reopen pr if you do.

PR #694, opened 2017-03-30
Animating state cleared when updating

Kronuz:
Tracker gets an outdated width when the callback arrives after an update, it gets the old value.


laveesingh:
No clear impact or the library has already included this functionality.

PR #697, opened 2017-04-04
cancel default behaviour for Drag and Drop and add active class .dragging when user dragging slide

Qunkero:
cancel default behaviour for Drag and Drop and add active class .dragging when user dragging slide


laveesingh:
I don't see a clear impact, or the library has already considered this point.

PR #651, opened 2017-02-22
Fix wrong if statement related with autoplay and resize

satazor:
This fixes the slider not autoplaying after a resize.

PR #710, opened 2017-04-10
Fix autoplay issue.

qr4:
I had an issue, that the slider won't slide automatically. I feel that the condition is either wrong, of the comment above it. Nevertheless, after I inverted the condition, the automatic sliding works as I expect it to.


qr4:
duplicate of https://github.com/akiran/react-slick/pull/651

PR #715, opened 2017-04-13
Fixes React 15.5 Warnings

wilbert-abreu:


wilbert-abreu:
@akiran Please merge this in as soon as possible as it's causing warnings on my end. Thanks!

erasmo-marin:
Great!! This will solve warnings but a migration to es6 classes is required in the future because importing createClass will increase bundle size for people that doesn't use it.

PR #719, opened 2017-04-19
Move enquire.js to dependencies

RocketPuppy:
Looks like b798b1fbd0aa50c73b3c0a2b506b7a27a3f40f31 introduced enquire.js, but added it to devDependencies instead of dependencies.

I noticed this when trying to build react-slick into my own project and the build started failing at 0.14.9.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,6 @@
     "css-loader": "^0.25.0",
     "deepmerge": "^1.1.0",
     "del": "^2.2.2",
-    "enquire.js": "^2.1.6",
     "enzyme": "^2.4.1",
     "es5-shim": "^4.5.9",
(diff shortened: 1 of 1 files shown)


RocketPuppy:
Looks like CI failed, but it looks like it's also failing on the latest master build https://travis-ci.org/akiran/react-slick/builds/223532995 (and this change shouldn't cause that failure anyway).

peterchoo:
It's worth noting that Travis CI did not pick up this issue as the test suite is run using `npm install`.  Perhaps there should be another test which runs on a clean slate with `npm install --production`.

akiran:
Can you check 0.14.10 version

PR #635, opened 2017-02-12
Fix for issue #628

nelsieborja:
Added explanation inside code for the change made

PR #723, opened 2017-04-19
Push correct slide number when lazy loading

### LINKED ISSUE #628: Last slide image not loading when swiping to last slide with enabled Lazyload
https://jsfiddle.net/20bumb4g/2113/

Swipe from left to right to scroll directly to the last slide, notice that image is not loading. It works fine when clicking on the prev button only when u swipe the issue is coming

tikotzky:
This commit will push the correct slide number on to the `lazyLoadedList` when swiping right on the first slide. Before this commit it would push `-1` instead of the correct slide number.

This PR is different than #635 in that it pushes the correct the index to the `lazyLoadedList`.
#635 just handles having the wrong index on the `lazyLoadedList`

fixes #628

### DIFF
--- a/src/mixins/helpers.js
+++ b/src/mixins/helpers.js
@@ -229,9 +229,10 @@ var helpers = {
       var loaded = true;
       var slidesToLoad = [];
       for (var i = targetSlide; i < targetSlide + this.props.slidesToShow; i++ ) {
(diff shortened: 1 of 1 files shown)


laveesingh:
This logic was updated in some commit in a previous release.

PR #754, opened 2017-05-12
changed children length to account for child arrays

zdenham:
hi, I really love this library! But I've been running into a bug where when the slider has more than two slides, and I perform slick next, on re-render, it will jump back to the second slide. After much hair-pulling, I realized it was because the `children.length` property was only 2 even though I had four child components. The first component was an array of length 3 (three slides) which were not accounted for. The reason for this was because I was using a map of three slides on the first element. i.e. 
```
<Slider ref={ref => this.slider=ref} {...settings}>

          {Tracks}

          <div>
            <DownloadDisplay hashtag={this.props.hashtag} />
          </div>

        </Slider>
```
 This change is just meant to account for maps of react elements... `let Tracks = keys.map((key, index) => {`.
Hope this helps! If not I would like to report the bug anyhow :)

### DIFF
--- a/src/inner-slider.js
+++ b/src/inner-slider.js
@@ -80,6 +80,14 @@ export var InnerSlider = createReactClass({
     }
   },
   componentWillReceiveProps: function(nextProps) {
+    
+    let length = nextProps.children.length;
+    for(var i = 0; i < nextProps.children.length; i++){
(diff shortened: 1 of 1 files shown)


akiran:
All the slides should be direct  children of Slider component.

We can't support nesting of slides because it could lead to lot of complex scenarios.

PR #755, opened 2017-05-15
Replace create-react-class usage with ES2015 class

ahutchings:
This lets us drop the create-react-class runtime dependency.

I added the babel-preset-stage-2 dev dependency so that we can use static class properties and property initializer syntax.


onnovisser:
Liking this, because this also makes React-Slick compatible with Preact.

ahutchings:
@akiran Is there something you'd like to see changed or updated before merging this?

PR #758, opened 2017-05-16
Prevent Carousel Sliding while drag-scrolling vertically on a touch device

### LINKED ISSUE #598: Carousel sliding while drag-scrolling vertically on a touch device
Link to original issue: https://github.com/kenwheeler/slick/pull/2521

In the link above you can see description of the bug. I think it's very important to fix that in react-slick too. So can we merge that changes in react-slick? I am ready to do that, if @akiran or somebody else accept this idea.

ryota-murakami:
Fixes #598

I fixed it for reference to PR of this original repository.(referenced on issue)
https://github.com/kenwheeler/slick/pull/2521/files

there is a difference in structure between the original and this library.(JQuery based, React based)
so i'm worry about whether this change is right.
especally following `if()` statement in the `swipeMove()` method.

original PR contain `_.scrolling` val in the condition.
https://github.com/kenwheeler/slick/pull/2521/files#diff-eaaa72f2e5f1d4974e4f1ae902717908R2674

but I did not make the equivalent change.
because when adding `this.state.scrolling` in following condition,
even when swiping only in the horizontal direction, it enters the following return statement and the slider stops.
https://github.com/akiran/react-slick/compare/master...ryota-murakami:fix/%23598?expand=1#diff-ae7d3bfc927feef9e44923133997daefR85

I'm welcome for feedback from everyone.

- [x] still test success.
- [x] no problem on browser behavior, as far as I was confirmed(launched by `npm start` sample page).


puja1234:
Still facing this issue, prevents autoplay in react-slick on touch devices while drag-scrolling vertically.
Followings are settings for react-slider

    dots: true,
    speed: 500,
    slidesToShow: 1,
    slidesToScroll: 1,
    autoplay:true

Version used is :-
"react-slick": "0.14.11"

laveesingh:
Please upgrade it to the latest version 0.22.2.

Issue #698, opened 2017-04-04
inconsistent animation for infinite scroll when lazyload enabled


dalimian:
normally, in all cases when lazyload is not enabled, clicking next when on last slide will animate "forward" to first slide as if the first slide repeats after last slide.

when lazyload is enabled, however, same action animates backward though all slides in between from last to first.

https://www.screencast.com/t/BtgPNqrU7

both animation styles are ok, but animation style should be a separate config option, independent of lazyload option.

apprears this code block is the cause for this inconsistent behaviour - https://www.screencast.com/t/6YBbG7CXr4H

rafaelderolez:
@akiran any word on this? I a fix for this has been merged before but I am still seeing this behaviour.

gkilian:
This issue looks to be fixed with this commit. https://github.com/akiran/react-slick/commit/7d8eeb5059a9f39931573913332f9785172f5a64#diff-ae7d3bfc927feef9e44923133997daef

PR #762, opened 2017-05-17
Infinite scroll animation bug

tplusrex:
## Cause

The problem was introduced with this line. Calculations doesn't support infinite scroll.
https://github.com/akiran/react-slick/blob/master/src/mixins/event-handlers.js#L27

## Effect

When lazy loading an infinite slide the last slide does not transition right instead it transitions left. You are moved to the first slide but for infinite galleries the first slide is appended to the right of  the last allowing for the infinite illusion to be seamless.

The issue is animated well here.
https://github.com/akiran/react-slick/issues/698

## Proposed Solution

Exclude infinite galleries from the condition. This allows infinite galleries with lazy loading to use the basic calculation.


dalimian:
@akiran really short pull request, can u merge this?

laveesingh:
Changes have been merged to dev branch, and soon will be merged to master.

PR #766, opened 2017-05-24
fix(dots): Fix dot count when slidesToScroll is smaller than slidesToShow

weaintplastic:
When configuring react-slick with slidesToScroll property smaller than slidesToShow the total number of dots shown is not accurate. 

This is now fixed and works for both, infinite and finite slider.

I've tested slider with the following setup with even (6) and odd (5) number of slides:

`slidesToShow={3} slidesToScroll={1} infinite={false}`
`slidesToShow={3} slidesToScroll={1} infinite`
`slidesToShow={1} slidesToScroll={1} infinite={false}`
`slidesToShow={1} slidesToScroll={1} infinite`
`slidesToShow={3} slidesToScroll={2} infinite={false}`
`slidesToShow={3} slidesToScroll={2} infinite`

Please let me know what you think. I'm happy to provide additional information whenever needed.

### DIFF
--- a/src/dots.js
+++ b/src/dots.js
@@ -3,9 +3,15 @@
 import React from 'react';
 import classnames from 'classnames';
 
-var getDotCount = function (spec) {
+var getDotCount = function getDotCount(spec) {
   var dots;
-  dots = Math.ceil(spec.slideCount / spec.slidesToScroll);
+
+  if (spec.infinite) {
+    dots = Math.ceil((spec.slideCount) / spec.slidesToScroll);
+  } else {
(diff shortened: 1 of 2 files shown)


weaintplastic:
@akiran do you have any thoughts on this. would love to get your feedback.

scolacur:
@akiran Is there a reason this hasn't been merged yet? The bug it addresses is fairly serious IMO.

laveesingh:
@scolacur @weaintplastic there is one major problem with this, `slidesToShow` and `inifinite` props are not accessible in `getDotCount` method, and it's breaking rendering and tests due to `slidesToShow` prop. 
Never mind, I'll fix it right away.

scolacur:
Thanks @laveesingh !

PR #873, opened 2017-10-02
Fixed warning when calling setState when the component has already be…

SBRK:
…en unmounted


laveesingh:
Could you please elaborate about what you are trying to achieve here? And please feel free to reopen the pr once you do.

PR #874, opened 2017-10-04
add isActive attribute on customPaging func

puemos:
If you need the style to dots and not by css.

```js
(nextIndex, isActive) => {
            const nextSocial = getNextSocial(nextIndex);
            return React.createElement(CardBody.svgMap[nextSocial], {
                  color: isActive ? 'red' : 'blue'
            });
        };
```


laveesingh:
@puemos, currently the prime goal of the library is to be as similar to slick as possible. So we try not to deviate from goal in any way and accept only features that slick possesses. Once we are close enough, we'll consider adding things that make it more convenient and flexible.

PR #890, opened 2017-10-18
Added afterRender method

silicakes:
The afterRender method method exposes the entire slider object, including state, props and calculated properties of all elements after each calculation or change.

type: function
parameters: sliderObject<ReactClass>

See newly added example and tests for details


laveesingh:
@silicakes currently the prime goal of the library is to be as similar to slick as possible. So we try not to deviate from goal in any way and accept only features that slick possesses. Once we are close enough, we'll consider adding things that make it more convenient and flexible.
Thanks a lot for the PR though, it might be helpful in the future.

PR #905, opened 2017-11-03
Add defaultProps in README

romainwn:
In order to make the documentation more clear, I thought that adding the defaultProps in the README could be a good idea, so people know what they don't need to call in the settings


akiran:
Merged this PR
https://github.com/akiran/react-slick/pull/942

Issue #240, opened 2016-02-18
Depends on jQuery


abhinavrastogi:
This adds 60kb to the webpack bundles. Highly inconvenient!
Would have gladly used it if it was much smaller.

akiran:
It doesn't depend on  jQuery.
It's purely written in react

yiminghe:
please move original slick to dev Dependencies

akiran:
It depends on slick for CSS.

yiminghe:
can you copy slick css to you project?

react-slick causes our build server to download unnecessary code (slick-carousel, jquery ...).

ShinyLeee:
I think slick-carousel is unnecessary too.. 
It depend on jquery and I have to download it.

Gervwyk:
Is there any way to remove the jquery dependency? Looks like it really adds unnecessary content to the bundle size.

akiran:
jQuery is never added in the bundle.  slick-carousel dependency is only for css

Gervwyk:
Ok. But that means jQuery ends up in bundle because of slick-carousel, which is only used for css. Can't we only copy the css to this project and remove slick-carousel dependency to get rid of jQuery in the bundle?

akiran:
We use wepack to bundle this library. We never imported jQuery. So it will not be bundled in this library.

Gervwyk:
Ok. Sorry. my bad. I missed the ⚠️ Also install slick-carousel for css and font part - which actually creates the issue. When using ant design and meteor, the import of antd causes it to import slick-carousel which is jquery dependent and causes this issue.

PR #911, opened 2017-11-10
add jquery to dependencies

DimaRGB:
because new version of slick requires jquery in peerDependencies.

https://github.com/kenwheeler/slick/releases/tag/1.8.0

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,12216 @@
+{
+  "name": "react-slick",
+  "version": "0.15.4",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "abab": {
+      "version": "1.0.4",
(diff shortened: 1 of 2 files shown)


harriha:
I'm not really fully up-to-date on how things work here, so please pardon me if I'm talking nonsense. However, if I understand correctly based on #240, the JS part of slick-carousel is not used and thus there is no real need for jquery. I just noticed the warning about missing peer dependency as well, but not sure what would be the best way to resolve this. Having jquery in the deps feels a bit wrong though if it's only for suppressing the warning and it is not being really used by the component.

DimaRGB:
@harriha sure, it's jQuery not used by react-slick, but until react-slick will depends on slick-carousel every project should add jquery to dependencies or only react-slick to prevent warnings.

akiran:
moved slick-carousel to peerDependency

PR #913, opened 2017-11-16
Add option to indent slider Netflix-like on left and right.

boxcee:
In my company we needed a Netflix-like slider that show next and previous slides.

This pull request extends the react-slick slider to allow indentation on both left and right.

Indentation will not work in vertical and infinite=false mode.

### DIFF
--- a/docs/demos.jsx
+++ b/docs/demos.jsx
@@ -1,32 +1,32 @@
 'use strict';
 
 import React from 'react';
-import Slider from '../src/slider';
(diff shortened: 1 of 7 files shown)


laveesingh:
We do not intend to have additional functionality that deviates from [slick](https://github.com/kenwheeler/slick) library. Still, if you disagree or have more to suggest, please free to reopen it.
Thanks for the PR though.

PR #893, opened 2017-10-23
fullscreen added

watzak:
This functionality provides fullscreen mode for images! 

screen shot 2017-10-23 at 15 08 53

screen shot 2017-10-23 at 15 09 02

Issue #870, opened 2017-09-29
fullscreen mode


watzak:
Hello guys!

Thanx for your library! But what I am missing is a **fullscreen mode**! Is there a possibility to do that anyway? Maybe I haven't seen it in the examples/documenation although I have checked them!

Thanx in advance for your feedback

Cheers watzak

watzak:
Here is a solution for fullscreen mode, what I have implemented =>
https://github.com/akiran/react-slick/pull/893

laveesingh:
We would love to have features like this in future. But currently, our prime goal is to be similar to and not to deviate from slick library. For now, I'm closing this, but once we are close enough to our goal, we'd definitely wanna have features like this.

watzak:
Hello @laveesingh

Is it possible to include 2 methods to your api! Variables still exist in your code!

Would be following methods:
`getDirection = () =>  this.innerSlider.getDirection();

  isAnimated = () => this.innerSlider.isAnimated();`

Thanx in advance for your feedback!

Cheers watzak

PS: Code still exist, but don't know if it's wanted for merging?

PR #916, opened 2017-11-17
Added Lightbox Mode

aaronholla:
Hi, 

I added a very basic Lightbox mode to the slider to be able to click on a slide and have it open in a popup box. This is very similar to #870 ( #893 ) adding a Fullscreen mode, except it doesn't force you into fullscreen it will instead give you a window with the slider in it. I added css classes for the lightbox to allow for further customization around overlay color and animating the opening of the lightbox.

| CSS Class | Description | lightbox
| ----- | ---- | ---- |
| .slick-wrapper | A wrapper div is always there even when lightbox is disabled | True/False
| .slick-lightbox-overlay | A container div for the lightbox | True
| .slick-lightbox | The actual div that will be the lightbox | True
| .slick-lightbox.open | Lightbox div will get an open class added when it is opened | True

Changelist
----

-Added "lightbox" prop to the slider 
-Added "lightboxOpen" state to the slider
-Added stopPropagation to all clickHandlers ( this prevents triggering the close lightbox )
-Added refs for wrapper & lightbox
-Added clickHandler to track slides for triggering lightbox if enabled
-Added example to the docs
-Added example styles for lightbox to function
-Added lightbox prop to the Readme

----

**Known bugs**
- When lightbox open and you maximize or minimize the window the slides are not the correct size and position. Resizing the window seems to fix this. Not sure what is causing this if it is a CSS issue or something I need to update.
- May have to add an update or something somewhere for AutoPlay to work properly 
- Wrapper trys to keep its size when lightbox is open. Noticed if there are margins on the slider that the wrapper ignores these margins and ends up smaller then it should be cause the HTML on the page to collapse by the size of the margins when lightbox is open
----
P.S.
_I could also expose a new method to toggle the lightbox. This would allow people to do things like create their own buttons to trigger the lightbox. Also would allow for a close button to be added by the user. ( think x in the corner if you wanted to style this like a fullscreen option )_

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,7 @@ class SimpleSlider extends React.Component {
 | easing         | string | | |
 | fade           | bool | Slides use fade for transition  | Yes |
 | focusOnSelect  | bool | Go to slide on click | Yes |
+| lightbox       | bool | Open a lightbox on click | Yes |
 | infinite       | bool | should the gallery wrap around its contents | Yes |
 | initialSlide   | int | which item should be the first to be displayed | Yes |
 | lazyLoad       | bool | Loads images or renders components on demands | Yes |
--- a/docs/_style.scss
+++ b/docs/_style.scss
@@ -30,6 +30,33 @@ h3 {
   margin: auto;
   width: 90%;
 }
+.slick-lightbox-overlay {
+  position: fixed;
+  top: 0;
+  left: 0;
+  z-index: 9999;
+  width: 100%;
+  height: 100%;
+  background-color: rgba(0, 0, 0, 0.85);
+}
+.slick-lightbox {
+  margin: auto;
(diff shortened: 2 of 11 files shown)


aaronholla:
Also for the tests I have to update all of the snapshots now because they will all get that new wrapper div.

I started some tests of the lightbox but haven't had time to fully build out the tests. I would like to add some tests to be sure that it plays well with other settings enabled.

akiran:
Lightbox is not supported by original slick library

We can't accept this PR to keep API compatibility with slick

aaronholla:
I understand. Thank you!

MihaiWill:
@aaronholla I am looking for this feature thanks! I will give it a try, I hope is still compatible

PR #924, opened 2017-11-25
fix, a moving in opposite direction on drag event at rtl option

ignocide:
i found problem at rtl option,
when drag a object, that moving in opposite direction,
you can check at example

### DIFF
--- a/src/mixins/event-handlers.js
+++ b/src/mixins/event-handlers.js
@@ -148,7 +148,12 @@ var EventHandlers = {
     }
 
     if (!this.props.vertical) {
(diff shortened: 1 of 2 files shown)


nimatrueway:
good job, but still incorrect behavior on first/last items when `infinite: true`.

laveesingh:
Thanks, @ignocide, and @nimatrueway 
That solves the mentioned problem, the pr will be merged Shortly

Linton-Samuel-Dawson:
I've installed react-slick with `npm` and still have this problem

caprica-Six:
Same here, even if direction is set to rtl: true, seems to render dir="ltr"...

PR #928, opened 2017-11-28
Fix prev button in center mode

sohai:
Prev button should work even though slideCount <= slidesToShow when center mode is enabled


laveesingh:
The prev button is already working as expected. Please, let me know if you disagree. 
Thanks for the PR, though.

PR #929, opened 2017-11-29
Add ability to specify tag name for track

VadimZhiltsov:
This would help to create SEO friendly markup.


laveesingh:
Keeping in mind that this library is just a port of slick library in jquery. We cannot support additional props. Thanks for the PR, though.

VadimZhiltsov:
@laveesingh yeah, but this library creates limitations with tag name.

My concern was about semanc markup. I'd like to make my slides as `<li>`, but because of current implementation of this library I have to place slides in `div`
Now we have something like this:
```
<div className='slick-track'>
    <li>Slide 1</li>
    <li>Slide 2</li>
</div>
```
What is not valid markup and has bad SEO affect. 

This is **not slick issue** and **not additional functionality**  , code change just removes restrictio created by this library.  Div limitation makes this library not perfect port for me

So I want to create ability to set any tag name to `.slick-track`

PR #933, opened 2017-12-04
Better check for IE8

strayiker:
window.addEventListener don`t work if we have a polyfill for other libraries as example


strayiker:
http://webreflection.blogspot.ru/2009/01/32-bytes-to-know-if-your-browser-is-ie.html

akiran:
React dropped support for IE8.
https://reactjs.org/blog/2016/01/12/discontinuing-ie8-support.html

We will also drop IE8 support and remove that code at some point.

Issue #764, opened 2017-05-19
Add .slick-active class on slides that are visible


weaintplastic:
When reaching the end of a finite slider with multiple slides, not all of them do get the `.slick-active` class.

Here's a little snapshot to describe the issue.

The slider is set up as following:
```
slidesToShow={4}
slidesToSroll={4}
infinite={false}
```

Initial - first 4 slides visible
```

<div class="slick-slide slick-active"></div>
<div class="slick-slide slick-active"></div>
<div class="slick-slide slick-active"></div>
<div class="slick-slide slick-active"></div>
<div class="slick-slide"></div>
<div class="slick-slide"></div>

```

After click on next arrow (actual behavoir) - last 4 slides visible
```
<div class="slick-slide"></div>
<div class="slick-slide"></div>
<div class="slick-slide"></div>
<div class="slick-slide"></div>
<div class="slick-slide slick-active"></div>
<div class="slick-slide slick-active"></div>
```

After click on next arrow (expeted behavoir)  - last 4 slides visible
```
<div class="slick-slide"></div>
<div class="slick-slide"></div>
<div class="slick-slide slick-active"></div>
<div class="slick-slide slick-active"></div>
<div class="slick-slide slick-active"></div>
<div class="slick-slide slick-active"></div>
```

Is this an expected bahvoir? If not, I've a fix ready and I'm happy to file a pull request.

Best

Roland

weaintplastic:
hey @akiran. did you have any chance to take a look at this proposal? Would love to get your feedback.

Thanks,

Roland

brownsmith:
You should be able to do this in your component using `componentWillReceiveProps` and the `nextProps`. You know the length of the array of slides, react-slick gives you your currentSlide, and you know how many slides you're showing in the carousel, in your case 4 slides.
If currentSlide = 0, and you're displaying 4 slides, then positions 0-3 in your array are 'active'

weaintplastic:
@brownsmith thanks for the fast reply. You mean the `componentWillReceiveProps` of my slide component? This is what I get there as `nextProps`:

```
children:Object
className:"slick-slide"
data-index:1
onClick:function onClick(e)
style:Object
tabIndex:"-1"
```
So there is no hint on the current slide included.

weaintplastic:
I'll try to hook into the `beforeChange` callback

weaintplastic:
[...]

PR #934, opened 2017-12-04
Fix bug on slickActive for items on last slide

marcbey:
Fix an issue with slickActive for items on last slide.
Same as https://github.com/akiran/react-slick/issues/764


laveesingh:
The tests are failing, please make sure you don't break other functionality while fixing a particular one. Please review and fix. Then we can open the PR again. 
Thanks for the PR, though.

PR #938, opened 2017-12-05
change bitwise OR to a logical OR

albertstill:
I presume this was a mistake because we have booleans either side of the OR? Cheers.


akiran:
Thanks for the PR @albertstill

PR #942, opened 2017-12-09
Update README with default props

JikkuJose:
Also updated a few descriptions & general styling.


akiran:
Thanks for the PR @JikkuJose

PR #943, opened 2017-12-11
Update react 16

wilbert-abreu:


dschinkel:
so does react-slick work with React 16?

PR #944, opened 2017-12-15
Scroll visible slides

yhidalgo:
Scrolling through all visible items, it works even when items have different widths. 
It introduces a new prop: scrollVisibleSlides, which is false by default, to preserve backward compatibility. 
I needed to add this functionality because slidesToScroll would scroll a static number of slides, while I wanted it to be dynamically calculated based on the width of the slides, and wether it was scrolling forward or backward.


laveesingh:
The changes that you are proposing, deviate from the original jquery slick library, so cannot be merged currently as we are trying to make it similar to the jquery one. Thanks a lot for the PR, though.

yhidalgo:
Thanks for contacting me! :)

The little bit of the jquery slick demos I checked out don't handle items
with different sizes:
https://www.jqueryscript.net/demo/Fully-Responsive-Flexible-jQuery-Carousel-Plugin-slick/

How should had I made it such that you can merge it?


Yoalli

 (\ /)
(  . .)
c('')('')

Issue #776, opened 2017-06-20
Server Side Rendering: Checksum invalid


rkkautsar:
```
Warning: React attempted to reuse markup in a container but the checksum was invalid. This generally means that you are using server rendering and the markup generated on the server was not what the client was expecting. React injected new markup to compensate which works but you have lost many of the benefits of server rendering. Instead, figure out why the markup being generated is different on the client or server:
 (client) ="slick-slide slick-cloned Testimony__Te
 (server) ="slick-slide slick-active slick-cloned 
```

unleashit:
I was having the same issue and have found that it comes from using the responsive property. If the current breakpoint doesn't match what's assigned to the slidesToShow property (the one that determines how many images are rendered on the server), you get the checksum mismatch. Since there's no easy way for the server to know the client's breakpoint (and therefore how many to render), ideally I think the client should at first respect the slidesToShow setting and then update it to the correct breakpoint on componentDidMount.

You can pretty easily do this yourself manually. Here's a bit of code that works for me. I'm taking the trouble to render a few slides on the server, but you could easily just render and empty string. Basically the script lazy updates the slider when the component mounts and sets the right number of images.

```
class FeaturedTattoos extends React.Component {

  constructor(props) {
    super(props);

    this.state = {
      clientRendered: false,
    };
  }

  componentDidMount() {
    if (isClient) this.setState({ clientRendered: true });
  }

  renderTattoos(tattoos, slidesToShow, server) {
    if (server && tattoos.length !== slidesToShow) tattoos.splice(4);

    return tattoos.map((tattoo) => {
      return (
        <div key={tattoo.filename}>
          <img
            src={`/tatteasyTattoos/${tattoo.filename}`}
            alt={tattoo.filename}
            style={{ maxHeight: '165px', height: 'auto' }}
          />
          <div className={style.getTattooButton}>
            <button className="btn btn-secondary btn-sm">Get Tattoo</button>
          </div>
        </div>
      );
    });
  }

  render() {
    const settings = {
      dots: false,
      infinite: true,
      speed: 500,
      slidesToShow: 4,
      slidesToScroll: 1,
      lazyLoad: false,
      arrows: true,
      responsive: [
        { breakpoint: 550, settings: { slidesToShow: 1 } },
        { breakpoint: 750, settings: { slidesToShow: 2 } },
        { breakpoint: 1000, settings: { slidesToShow: 3 } },
        { breakpoint: 1600, settings: { slidesToShow: 4 } },
[...]

PR #950, opened 2017-12-23
Add options for server side rendering responsive sliders

techwes:
This is a PR to close issue #776. I have added an `ssr` prop that delays the setting up the responsive handlers until `componentDidMount`. This means the slider will render with default settings on the server. The client then renders also with the default settings so there is no checksum error and then is updated with the correct responsive settings and immediately re-rendered. I also added the ability to set which breakpoint settings the slider should use to render on the server side. This is useful if you have server side device detection and can guess which breakpoint your client will use. Also updated the docs to explain the new props.

The changes ended up being minimal and I have tested this within my project. Tried to write a test but found it very difficult since I needed to assert something was the case before `componentDidMount` and then also mock the enquire.js dependency. But happy to still write one with some guidance on how.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,8 @@ Props            | Type            | Default Value                   | Descripti
 `slidesToShow`   | `int`           | `1`                             | Yes                                                         | Yes
 `slidesToScroll` | `int`           | `1`                             |                                                             |
 `speed`          | `int`           | `500`                           |                                                             |
+`ssr`            | `bool`          | `false`                         | True if you SSR and use `responsive` prop (see below)       | Yes
+`ssrBreakpoint`  | `int`           | `null`                          | The breakpoint to use during SSR (see `responsive` below)   | Yes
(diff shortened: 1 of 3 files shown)


laveesingh:
Your changes propose adding a new prop, the library doesn't support props that deviate from [slick](https://github.com/kenwheeler/slick/) library.

techwes:
Okay I understand wanting to keep the API the same, but since this is **react**-slick and **react** supports server side rendering this is still an issue. I added a prop to maintain the existing behavior for people who are no doing ssr. But I could just change the behavior so the responsive handlers are always set up in `componentDidMount` which honestly still makes sense. It will just mean an extra render for client side rendered sliders when the screen is within a breakpoint. Does that sound good to you?

A way to set the initial breakpoint (which I called `ssrBreakpoint`) would still be great but is not necessary for my current use case. Are there no plans to ever have props not in slick? There could be a prop called `reactOptions` that holds all of these and makes it clear what is a slick property and what is specific for react-slick

PR #959, opened 2018-01-08
fix when use lazyLoad in centerMode the elements on the left side can…

luckyadam:
fix when use lazyLoad in centerMode the elements on the left side can not be rendered
wx20180108-160919 2x

### DIFF
--- a/src/inner-slider.js
+++ b/src/inner-slider.js
@@ -39,9 +39,26 @@ export var InnerSlider = createReactClass({
       mounted: true
     });
     var lazyLoadedList = [];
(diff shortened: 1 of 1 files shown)


laveesingh:
The PR fixes the said issue and will be merged shortly. 
But, did you notice that there is a similar issue without centerMode?
Try these configs to regenerate:
`const settings = {
      dots: true,
      lazyLoad: true,
      infinite: true,
      speed: 500,
      slidesToShow: 3,
      slidesToScroll: 1,
      initialSlide: 2,
      // centerMode: true,
    };`

PR #960, opened 2018-01-08
Remove slick-carousel dependency

vio:
Avoid getting the npm warning regarding slick-carousel peer dependencies (jquery)


canda:
And how do you add the `slick-carousel` css files?
I installed `slick-carousel` as a dependency on my project and I'm getting a webpack error saying that it can not load jquery.
How did you work around this problem @vio @akiran ?

PR #964, opened 2018-01-11
Center mode fix

sajjad26:
in case we have slidesToShow is greater then 1, 
I have also updated the example page CenterMode.js to show the result.

### DIFF
--- a/examples/CenterMode.js
+++ b/examples/CenterMode.js
@@ -14,7 +14,7 @@ export default class CenterMode extends Component {
     };
     return (
       <div>
-        <h2>Center Mode</h2>
+        <h2>Center Mode Finite</h2>
         <Slider {...settings}>
           <div><h3>1</h3></div>
           <div><h3>2</h3></div>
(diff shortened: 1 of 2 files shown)


laveesingh:
What exactly are you trying to fix here, could you tell more details?

laveesingh:
One thing to take note of here is that we're going to **enforce scrolling by one when the center mode is enabled**. This is the behavior that the original slick library shows and we don't want to deviate from that. So the issue that your PR fixes, will get fixed automatically. Thanks for the PR, though.

sajjad26:
Hi,
The issue emerges when you have this kind of configuration
```javascript
var settings = {
    dots: true,
    slidesToShow: 3,
    slidesToScroll: 3,
    centerMode: true,
    infinite: false,
    centerPadding: '20px'
};
```
Example is show here http://jsfiddle.net/2pbkhaxr/
I needed this to work so I figured others might as well.
Anyway great library though.
Thanks

PR #980, opened 2018-01-23
Add slickListClasses as a Prop

Yamikamisama:
Adds a new prop slickListClasses to the Slider allowing you to pass custom CSS classes to the slick-list wrapper.

Very Useful if you are trying to customize the carousel's styles

Also updated the readme to reflect the change

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,47 +81,48 @@ class SimpleSlider extends React.Component {
 
 ### Props
 
-Props            | Type            | Default Value                   | Description                                                 | Working
----------------- | --------------- | ------------------------------- | -----------                                                 | -------
-`accessibility`  | `bool`          | `true`                          | Enable tabbing and arrow key navigation                     | Yes
-`className`      | `string`        | `''`                            | CSS class for inner slider div                              | Yes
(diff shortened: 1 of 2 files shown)


laveesingh:
We already have a `className` prop that fulfils the purpose. Thanks for the PR though. If you disagree or have more to suggest, please feel free to reopen the PR.

Yamikamisama:
@laveesingh right I know that className exists that is for the parent div (top of the Component), which lets me do basic styling around how the outer shell looks. However using CSS modules wont allow me to access the inner div containing `react-slick` without doing some ugly CSS magic since it creates a namespace of `brand---filename---classname` therefore never being === to `react-slick`.

```
> div:first-of-type {
    margin: 0 84px;
}```

or using `:global`

both of which are less than ideal as I would like them to act indecently and leave no room for accidentally overwriting the classes in other global files. I still propose that we add this in allowing us to have more control of the inner div

laveesingh:
@Yamikamisama currently prime goal of the library is to be as similar to [slick](https://github.com/kenwheeler/slick) as possible. So we try not to deviate from goal in any way and accept only features that slick possesses. Although once we are close enough, we'll consider adding things that make it more convenient and flexible. 
Thanks a lot for the PR though, it might be helpful in the future.

PR #990, opened 2018-01-30
Added current slide identity class

WinfredMa:
Reason: I can't set current slide style under multiple items mode.
Update: Added 'current-index' to current slide when has multiple active slides.


laveesingh:
The problem has been taken into account, but it needs to be solved in a different way, we'll try and do that. Thanks for the PR though.

WinfredMa:
Thanks.  I'm looking forward to the solution.

PR #1002, opened 2018-02-04
Add afterDots prop for an element after the props

NiQDeveloper:
The project I am working on requires an additional element after the dots have been render.

This would give a user a simple hook for rendering a custom element in the slider (eg. A zoom button, reset button, switch etc)

### DIFF
--- a/src/default-props.js
+++ b/src/default-props.js
@@ -46,7 +46,8 @@ var defaultProps = {
     swipeEvent: null,
     // nextArrow, prevArrow should react componets
     nextArrow: null,
-    prevArrow: null
+    prevArrow: null,
+    afterDots: null
 };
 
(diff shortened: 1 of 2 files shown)


laveesingh:
Individual changes that deviate from [slick](https://github.com/kenwheeler/slick) for no significant impact, can't be merged.
Thanks for the PR though.

NiQDeveloper:
Hi @laveesingh 

The functionality from slick that I am missing/would normally use to fix this solution is "appendDots".

This defines a container for where the dots are generated.

I propose adding a new attribute similar to customPaging, to handle the dots container, with the default prop being the current method of rendering dots.

I will open another PR with the proposed functionality for your review.

PR #1009, opened 2018-02-07
Add renderDots for custom rendering the dots

NiQDeveloper:
Following on from [PR 1002](https://github.com/akiran/react-slick/pull/1002). This would add an attribute similar to that of appendDots in [slick](https://github.com/kenwheeler/slick).

It is done in the same style as `customPaging`, where users can provide a function for rendering the dots within a custom container. Meaning it should be both inline with the slick library, as well as inline with the coding standards of react-slick.

Thank-you for your consideration.


stephane-r:
@akiran do you think we can merge it ? I think it's great idea :)

paulcredmond:
A demo on using this would be great.

Issue #1010, opened 2018-02-08
Slider doesn't update width without triggering component rerender.


GeorgySerga:
The issue is the same as here pretty much: https://github.com/akiran/react-slick/issues/809 ,- without `resize` event, or any other trigger for components to rerender (modification of the props for example) the images inside slider do not adjust their size. Here is an example:
![2](https://user-images.githubusercontent.com/9346874/35949072-be200fa2-0c3d-11e8-90e6-aef4e9afc805.gif)

I took a look into this and was able to fix this using css `calc()` property, will PR soon.

GeorgySerga:
This one should also fix all of these: https://github.com/akiran/react-slick/issues/600, https://github.com/akiran/react-slick/issues/543 .

laveesingh:
Just add a ref in wrapper:

```
<Slider {...settings} ref={slider => this.slider = slider}> ... </Slider>
```
and doing `this.slider.innerSlider.onWindowResized()` after setting states in wrapper, should take care of the updates.

laveesingh:
Will try to come up with a better js solution for this.

laveesingh:
https://github.com/akiran/react-slick/commit/8591dace983fa1bc103964459ab99eaa34eed8de solves the problem. Changes will be released soon.

PR #1011, opened 2018-02-08
Make slider and images width responsive with calc(), see #1010

GeorgySerga:
https://github.com/akiran/react-slick/issues/1010

### DIFF
--- a/examples/__tests__/__snapshots__/SimpleSlider.test.js.snap
+++ b/examples/__tests__/__snapshots__/SimpleSlider.test.js.snap
@@ -6,43 +6,43 @@ exports[`Simple Slider Snapshots click on 3rd dot 1`] = `
     <div class=\\"slick-initialized slick-slider\\"><button type=\\"button\\" data-role=\\"none\\" class=\\"slick-arrow slick-prev\\" style=\\"display: block;\\"> Previous</button>
         <div class=\\"slick-list\\">
             <div class=\\"slick-track\\" style=\\"opacity: 1;\\">
-                <div data-index=\\"-1\\" class=\\"slick-slide slick-cloned\\" style=\\"width: 0px;\\">
+                <div data-index=\\"-1\\" class=\\"slick-slide slick-cloned\\">
                     <h3>6</h3>
                 </div>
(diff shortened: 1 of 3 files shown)


GeorgySerga:
Hi @laveesingh , I know you can review and merge in this repo, would you please find some time and take a look. Appreciate it, thanks!

laveesingh:
@GeorgySerga sure asap!

laveesingh:
@GeorgySerga I don't think this css solution to the problem will be a good idea. Because if the component doesn't update it's going to create several problems especially with trackLeft. I think forcing component update in the wrapper will be a good idea.
Just add a ref in wrapper:
```
<Slider {...settings} ref={slider => this.slider = slider}> ... </Slider>
```
and doing `this.slider.innerSlider.onWindowResized()` after setting states in wrapper, should take care of the updates.

GeorgySerga:
@laveesingh Hi, thank you for your answer and review. I don't think we understand the problem here the same way. What I'm trying to solve is slider resize in case of no window resize. Some elements on the page open or close, but window resize event wouldn't fire. I haven't seen any problems with this solution, if it doesn't solve all the problems when window size isn't changed, but solve some of them - it's still a solution, I had none of the other issues so far. I'll add a couple of extra gifs to illustrate the point shortly.

GeorgySerga:
Though I think I see what you mean about trackLeft issue, I should also update `transform3D`.

laveesingh:
@GeorgySerga our ultimate plan is to attach a listener to slider resize instead of entire window resize, but we're struggling with that a bit. The problem with pure CSS solution is that we're using calculated values to determine a lot of factors such as trackWidth, listWidth, slideWidth, trackLeft etc and also these values help with consequent results there and if the values aren't correctly updated we may not get desirable visual.
I'd welcome a JS solution.
In fact, I'm reopening the #1010 just to come up with the solution myself.
Thanks for your efforts, though.

sag1v:
>  our ultimate plan is to attach a listener to slider resize instead of entire window resize, but we're struggling with that a bit.

@laveesingh I've done it in #1212 it is waiting for a review (2 months now) :)

Issue #429, opened 2016-08-22
Adding a post init callback inside inner-slide


gkilian:
Currently we're running  `initialize()` and then `adaptHeight()` in the `componentDidMount` of `inner-slide.jsx`. 

**Issue:**  For a second or two, while we are adapting the height of the `inner-slide`, the component's height is 100% the height of the stack of slides and then they snap into place.

![images-stacked](https://cloud.githubusercontent.com/assets/801136/17866089/38cf139e-6873-11e6-80dd-83b0e3e40d3c.png)

**Proposed solutions**
1. Add a `postInit` callback after `adaptHeight`, so the user can do what they need after (_eg. show the slider_)
2. Or add an initialized class to the component after we adapt the height, similar to what we do with `slick-slider`.

nmynarcik:
Can the CN fix referenced above be applied to this project?

gkilian:
@nmynarcik - It can! It's the first proposed solution above.

nmynarcik:
this has been a hurdle for me, since I have this lib inside of a ReactTransitionGroup and it prevents proper animation since the DOM is being adjusted during the parent transition. So I need a way to tap into after its been initiated so I can fade it in.

gkilian:
@nmynarcik - Sorry for the long wait. I have some time to add PRs to this repo, so hopefully this works and is accepted. 

https://github.com/akiran/react-slick/pull/1013

laveesingh:
Can't support props that deviate from slick.
If you want this issue open please regenerate the problem on codesandbox.

gkilian:
@laveesingh - The only other approach that doesn't require an additional prop would be to add a classname, similar to what `initialize()` does to `adaptHeight()`.

PR #1013, opened 2018-02-13
Add afterInit function after init and adaptHeight

gkilian:
This PR address the issue found [here](https://github.com/akiran/react-slick/issues/429), regarding sudden flash prior to inner-slider loading.

We've have added this to Conde's internal fork and have seen great success!


laveesingh:
Can't support props that deviate from [slick](https://github.com/kenwheeler/slick/).
Thanks for the PR though.

PR #1016, opened 2018-02-15
Fix responsive behavior. Issue #777

### LINKED ISSUE #777: Responsive breakpoints are overlapping and causing undesired behavior
The breakpoints in the responsive array generate overlapping media queries.
In this example https://github.com/akiran/react-slick/blob/master/examples/Responsive.js the breakpoints are 480, 600 and 1024. If you open a browser with 480 screen width (or adjust it via devtools) and refresh the page you can see two items per slide in the "Responsive" example while there should be only 1, according to the breakpoints. 
I'm guessing this happens because the carousel settings are controlled by matchmedia listeners and both listeners are getting called. This would be expected since the media queries generated by https://github.com/akiran/react-slick/blob/master/src/slider.js#L32 would overlap. A possible solution would be to increase the lower boundary (min-width) of the match medias by 1 so that only one matchmedia listener would be called for a given screen size.

vkrbt:
This code fix #777 issue. Now it works like `max-width` media query described in readme.

### DIFF
--- a/src/slider.js
+++ b/src/slider.js
@@ -37,7 +37,7 @@ export default class Slider extends React.Component {
         if (index === 0) {
           bQuery = json2mq({minWidth: 0, maxWidth: breakpoint});
         } else {
-          bQuery = json2mq({minWidth: breakpoints[index-1], maxWidth: breakpoint});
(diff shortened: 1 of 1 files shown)


laveesingh:
This has been merged to dev, will be released to master soon.

Issue #556, opened 2016-11-23
Responsive / slideshow width update


jbracoupeau:
Hi guys, I'm using this awesome slideshow on one of the websites I'm working on.

The only problem I have is that when I resize the window, the slideshow also resizes (as the .slideshow div has width:100%), but the .slick-slide .item dev gets for example width:500.5px (which is the .slick-track width/4)... 
Everytime the .slick-slide .item's width is not a round number, we can see 2 pixels of the next pic on the right side, any parameter to change this behaviour ? I would like to Maths.ceil() it + 1px ;)

Cheers guys.

dijonkitchen:
Related to #221?

jbracoupeau:
yes, I tried to update to 0.14.5 this morning, but it's still there right ?

PR #1020, opened 2018-02-19
merge dev to master

laveesingh:
**In progress**

Following are the changes to be mentioned:

- fixed slideWidth calculation approximation
- fixed unusual scrolls in focusOnSelect mode
- added appendDots method for customization of dots
- modified logic for handling odd/even cases where there were unusual scrolls in opposite direction
- implemented unslick feature properly
- fixed variableWidth issues like blank spaces at edges, improper alignment
- handling focus loss in case of fade=true
- responsive lazyloading bug fixed
- increased verticalswiping resistance from 4 to 10

### DIFF
--- a/src/dots.js
+++ b/src/dots.js
@@ -59,10 +59,6 @@ export class Dots extends React.Component {
       );
     });
 
-    return (
-      <ul className={this.props.dotsClass} style={{display: 'block'}}>
-        {dots}
-      </ul>
-    );
+    return React.cloneElement(this.props.renderDots(dots), {className: this.props.dotsClass});
   }
 }
--- a/src/dots.js
+++ b/src/dots.js
@@ -59,6 +59,6 @@ export class Dots extends React.Component {
       );
     });
 
-    return React.cloneElement(this.props.renderDots(dots), {className: this.props.dotsClass});
+    return React.cloneElement(this.props.appendDots(dots), {className: this.props.dotsClass});
   }
 }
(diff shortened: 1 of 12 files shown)


src/dots.js
```diff
@@ -59,10 +59,6 @@ export class Dots extends React.Component {
       );
     });
 
-    return (
-      <ul className={this.props.dotsClass} style={{display: 'block'}}>
-        {dots}
-      </ul>
-    );
+    return React.cloneElement(this.props.appendDots(dots), {className: this.props.dotsClass});
```
akiran: @laveesingh Do we need cloneElement here ?

We can do some thing like this
```
const AppendDots = this.props.appendDots
return <AppendDots className={this.props.dotsClass} />
```

laveesingh: I'll try and replace that.

src/mixins/event-handlers.js
```diff
@@ -301,21 +316,22 @@ var EventHandlers = {
...
+          this.setState({ currentDirection: 0 }) // unverified fix of above line
           break;
 
         case 'right':
         case 'up':
           newSlide = this.state.currentSlide - this.getSlideCount();
           slideCount = this.props.swipeToSlide ? this.checkNavigable(newSlide) : newSlide;
-          this.state.currentDirection = 1;
+          // this.state.currentDirection = 1; // critical: change this line with setState statement
+          this.setState({ currentDirection: 1 }) // unverified fix of above line
```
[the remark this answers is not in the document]
laveesingh: absolutely

dijonkitchen:
FYI, this likely does the following:

Fixes #94
Fixes #221
Fixes #556

PR #1070, opened 2018-02-27
Update trackHelper.js

SrWildman:
If we are looking at the first slide there is no need for offset regardless of how many slides there are. Keep everything without translation.

### DIFF
--- a/src/mixins/trackHelper.js
+++ b/src/mixins/trackHelper.js
@@ -92,7 +92,7 @@ export var getTrackLeft = function (spec) {
   if(infinite){
     slidesToOffset = -getPreClones(spec) // bring active slide to the beginning of visual area
     // if next scroll doesn't have enough children, just reach till the end of original slides instead of shifting slidesToScroll children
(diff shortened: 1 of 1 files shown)


laveesingh:
Could you talk a bit more about what visual inconsistency are we trying to fix here? Perhaps, a codesandbox demo would be quite useful.

laveesingh:
Could you talk a bit more about what visual inconsistency are we trying to fix here? Perhaps, a codesandbox demo would be quite useful.

SrWildman:
@laveesingh, Sorry for not providing more information the first time. Below is a playground demonstrating the issue. When you have fewer cards than are intended to be displayed, they are right aligned. I used the settings in our project that reproduce this issue.

https://codesandbox.io/s/2pk30x04zn

laveesingh:
Thanks for pointing out the issue and putting efforts to solve it. But the reference commit 8339b67 does a better job of handling it.
Thanks for the PR though.

SrWildman:
Awesome, thank you.

Issue #1082, opened 2018-03-02
Positioning is broken when [variableWidth = true], [infinite = false] and [centerMode=false]


patrickvienneau:
[Sandbox example](https://codesandbox.io/s/ll8o1lk0wm)

Seems that when the following settings are set, the initial positioning isn't correct (still centered), while the positioning when cycling through slides is entirely broken:

```
centerMode: false,
infinite: false,
variableWidth: true,
```

### Expected Behaviour
On initialization, I would expect the first slide to be positioned to the left of the container. On changing of slides, the active slide should align itself to the left as well.

oleg-ch:
It is not present in version `^0.15.4`, so untill the issue is resolved, you may use that version as a temporary alternative. 

The calculation of initial left positiion has been changed quite a bit since 0.15.4 ([diff compared to 0.19](https://github.com/akiran/react-slick/compare/8f1d7a8...4908787#diff-281728d6ebdeaa3f5cb589407711cb2cL132)).

Also there's already a [PR](https://github.com/akiran/react-slick/pull/1083)

patrickvienneau:
Thanks @oleg-ch I'll look into that version, thanks for following up.

laveesingh:
8b8b42c67704e855f1f907e9e256f9f8d0e75e8b

patrickvienneau:
@laveesingh See https://github.com/akiran/react-slick/commit/8b8b42c67704e855f1f907e9e256f9f8d0e75e8b#r27930485 as there appears to be some unintended side effects from your commit.

laveesingh:
I think that is intended effect. Following jquery slick demo reflects similar behavior: https://codesandbox.io/s/61wyv4xlr

patrickvienneau:
Sure looks like it, thanks for clarifying that up!

PR #1083, opened 2018-03-02
Fixes Positioning of Slides for [variableWidth=false][centerMode=false][infinite=false]

pvienneau:
See #1082 for context.

## Changes
- Only compensate `targetSlideIndex` with `preClones` if `spec.infinite=true`
- Only correct `targetLeft` if `targetSlideIndex > 0`; `targetLeft` should be of `0` for the first slide.


laveesingh:
The PR doesn't actually fix the problem. Single swipe recreates the problem. Though, I think I have managed to resolve it. 8b8b42c67704e855f1f907e9e256f9f8d0e75e8b

PR #1089, opened 2018-03-07
Correct contributors link in CONTRIBUTING.md

joehua87:
Correct contributors link


laveesingh:
What effective different does it make?

PR #1093, opened 2018-03-08
Remove slide style when "unslick" setting is active.

DenisFrezzato:


laveesingh:
I don't think we're gonna do that. Slick carousel reflects the same behavior so we're gonna stick with it.

PR #1107, opened 2018-03-15
Responsive on DOM element size

vitalynait:
+ Property to track the size of a particular element in a DOM tree


laveesingh:
Thanks for pointing out the problem. But we need to merge it to dev before it goes to master and as you can see dev branch has changed the entire InnerSlider component, so it's not possible for me to merge the PR. Though the commit: 8591dace983fa1bc103964459ab99eaa34eed8de does something similar.

PR #1122, opened 2018-03-21
Added ability to make the center slide width grow and shrink

miketamis:
Checkout the demo to understand more


laveesingh:
It certainly looks good, but [slick](https://github.com/kenwheeler/slick) doesn't support any such functionality and we are strictly trying to stick to it.
Thanks for the efforts, though.

Issue #1127, opened 2018-03-22
None clear setTimeout may cause error


popo1221:
Hi @akiran ,

I found the code on [line 111 of inner-slider.js](https://github.com/akiran/react-slick/blob/75ce01b0048ceee4d822bdb5c30f3dbfe14a46c3/src/inner-slider.js#L111)
```js
        image.onload = () => setTimeout(() => this.update(this.props), this.props.speed)
```
cause errors. If the component destroyed before the images loaded.

![image](https://user-images.githubusercontent.com/4590675/37748705-d0d71426-2dbf-11e8-9e37-3a078946ecf2.png)

laveesingh:
f704b92477023a89813d1593da3821d834673ff7 takes care of it.

PR #1128, opened 2018-03-22
Clear image onload callbacks after component unmount #1127

popo1221:


laveesingh:
Thanks for pointing out the problem. But we need to merge it to dev before it goes to master and as you can see dev branch has changed the entire InnerSlider component, so it's not possible for me to merge the PR. Though the commit: f704b92 does something similar.

PR #1158, opened 2018-03-30
merge dev to master

laveesingh:
**Changes to mention**

- Bug fixes
  - fixed initial render flicker in server side rendering
  - fixed breaking animation due to resize observer updates
  - fixed image on click focus on slider
  - fixed IE incompatibility with forEach
  - improved performance by avoiding unnecessary updates by using pure components
  - added clickHandler to avoid clicking while swiping
  - fixed vertical body scrolling while vertical swiping
  - improved performance by debouncing onWindowResize
  - fixed dynamic animation change break

- Features
  - added rows and slidesPerRow functionality
  - added aria-hidden property to slides
  - added dontAnimate argument to slickGoTo method

- Other changes
  - added eslint tests
  - added script for jquery demo generation
  - removed redundant classes after implementing different kinds of unslick behavior
  - adding slick-initialized class on client and not on server in case of server side rendering
  - removed props table from readme and added a link to the documentation instead


blai-cartera:
Any idea on merge date? Some critical bug fixes in here :) Thanks for the great work!

laveesingh:
@blai-cartera it should be released today.

PR #1178, opened 2018-04-08
Allow changing responsive prop after mount

danilobuerger:
Helpful when you render server side with something like react-snap. It will create the slider with a certain breakpoint configuration which needs to be hydrated (react 16 won't patch diff). This way we can always render in a predetermined breakpoint setting and apply the rest on the client on a second render pass.


danilobuerger:
ping @akiran

danilobuerger:
not stale

akiran:
@maddhruv Yes, this issue need to addressed.

PR #1179, opened 2018-04-09
Dev

akiran:


yenly:
Thank for the quick fix and response to my bug reporting.  This one is a critical bug for me. I'm curious to when this PR would be merged? Thanks in advance for all your hard work!

PR #1199, opened 2018-04-20
use existing child key if present

nico1510:
The current logic assumes that users did not specify unique keys for the slides which are children of the Slider component so it generates them manually using indices. I'm using dynamic slides/children however and this leads to problems when I want to unmount a slide at say index position 0 because all subsequent slides will have their index changed and hence have to update. By using the child key (if present) this can be avoided.

### DIFF
--- a/src/slider.js
+++ b/src/slider.js
@@ -166,6 +166,7 @@ export default class Slider extends React.Component {
       i += settings.rows * settings.slidesPerRow
     ) {
       let newSlide = [];
+      const childKey = children[i].key;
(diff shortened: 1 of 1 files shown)


gabrielloeb:
Would love to see this merged!

margaritachepikova:
any chances to have this fix merged?

aedenmurray:
Any updates on this?

DmytroKukharenko-q4inc:
@akiran any chance to merge this fix?

PR #1242, opened 2018-05-23
fixed variableWidth slides overflowing to multiple lines

### LINKED ISSUE #1241: Slides going to multiple lines with variableWidth: true
### Problem

When using `variableWidth: true` the track width is miscalculated when one (or more) of the slides are wider than the container. This results in the track contents overflowing to multiple lines (and the slider becoming very high).

Original slick tackles this by setting the slideWidth always to 5000px when variableWidth is used (cf. https://github.com/kenwheeler/slick/blob/a2aa3fec335c50aceb58f6ef6d22df8e5f3238e1/slick/slick.js#L2081)

### Example from examples

This is visible in this example, when you resize the window: https://react-slick.neostack.com/docs/example/variable-width

screen shot 2018-05-23 at 14 59 38

apkoponen:
Fixes #1241

### DIFF
--- a/src/utils/innerSliderUtils.js
+++ b/src/utils/innerSliderUtils.js
@@ -567,7 +567,11 @@ export const getTrackCSS = spec => {
   let trackWidth, trackHeight;
   const trackChildren = spec.slideCount + 2 * spec.slidesToShow;
   if (!spec.vertical) {
-    trackWidth = getTotalSlides(spec) * spec.slideWidth;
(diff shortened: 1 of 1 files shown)


lorenzos:
What if the slide is bigger than 5000px? Isn't there a way to get the correct value for `spec.slideWidth` in the beginning, independently from windows size?

wcerfgba:
@akiran Any movement on this?

PR #1270, opened 2018-06-08
Pass in if current dot is activeDot on customPaging

browniefed:
Sometimes having access to the active dot is necessary in the customPaging callback


LuigiClaudio:
That would be great to have, any updates on this pull request? :)

browniefed:
No one cared to merge it sooo not clue.

PR #1291, opened 2018-06-28
explicitly setting to be not-passive to fix iOS 11.3 issue

HelKyle:
https://bugs.webkit.org/show_bug.cgi?id=184250


nikhilatjumpcut:
hey @akiran, can we merge this PR if it's ready?

This is kind of a similar thing that happens on Chrome v56.0+
- https://www.chromestatus.com/features/5093566007214080
- https://developers.google.com/web/updates/2017/01/scrolling-intervention

NeryC:
I had this problem and this solution worked very well for me, you should merge this

PR #1294, opened 2018-06-28
Preserving custom style on slider

HelKyle:


thiennhan2310:
@akiran  When will this be merged?

PR #1297, opened 2018-07-01
option to pre-load left right images

nidhi-ag:
When image size is large, it is a good user experience to pre-load next image.


sweber83:
We are very interested in this feature.
Any chance we could get this merged soon?
Anything we can do to speed up the process?

PR #1308, opened 2018-07-11
Adjust track width calculation for variable width slides

### LINKED ISSUE #1304: variableWidth and left issue
Hi guys,

My carousel is "full screen" meaning that it is pushed to the left and top with width 100% + pushed difference width and position absolute. Pushed from its original place. In order to have slides working properly I use `variableWidth: true`, but value is not taken into consideration when it calcs left offset.

**Current behavior**:
With `variableWidth: true` and width of container 150 it slides to 130 left because it is original width of container not that, which is set by `variableWidth`.

**Expected behavior**:
With `variableWidth: true` and width of container 150 it slides to 150 taking into account container width, which is set by `variableWidth`.

randallreedjr:
Fixes #1304 

For variableWidth=true, the typical calculation of trackWidth does not work (images do not necessarily take up an equal amount of the viewable area in the carousel). Slick Carousel gets around this by using an arbitrarily large slideWidth (5000).

https://github.com/kenwheeler/slick/blob/master/slick/slick.js#L2081

### DIFF
--- a/src/utils/innerSliderUtils.js
+++ b/src/utils/innerSliderUtils.js
@@ -48,7 +48,7 @@ export const getSwipeDirection = (touchObject, verticalSwiping = false) => {
   xDist = touchObject.startX - touchObject.curX;
   yDist = touchObject.startY - touchObject.curY;
   r = Math.atan2(yDist, xDist);
-  swipeAngle = Math.round(r * 180 / Math.PI);
+  swipeAngle = Math.round((r * 180) / Math.PI);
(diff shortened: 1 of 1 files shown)


MickeyKay:
@randallreedjr @YIZHUANG any update on this? If this fix works I'd love to see it implemented.

randallreedjr:
@MickeyKay Not sure what update I can provide. I made the fix, it worked for me, I opened the PR, it was closed. Over a year later, we're using the code from my fork.

PR #1337, opened 2018-08-10
Finite variable width slider slides always fill the display area

### LINKED ISSUE #1334: Finite variable width slider (that scrolls only visible slides), slides should always fill the display area
>I hope this is not a "big ask" but I'm stuck and I love your library as it has helped me a lot in the past.

## Overview
My issue is that I'm trying to adjust the [variable width](https://react-slick.neostack.com/docs/example/variable-width/) slider (in your documentation) in order to be:
1. `finite`
2. `responsive` (to different viewports) (if possible to display as many slides according to viewport width)
3. `scroll only 100% visible slides`
4. slides should fill the `div` display area

### Observations

1. So far, I have managed to make it:
   - `finite`

2. Regarding `responsive`, breakpoints don't work well for `variable width` slides.

3. `scroll only 100% visible slides`, This is very important when it comes to variable viewport width and when using variable width slides.

4. `slides should fill the`**div**`display area`, _finite variable width slides_ when they reach the last slide, even though it was already visible by the previous scroll, the user can still scroll and what is visible is a huge area of blank slides, for example:
blank slides visible after last slide

## The Code
The issue can be replicated and tested in [code sandbox](https://codesandbox.io/s/983v21ooq4)

> I'm open to hear your expert solutions.

diegopamio:
Fixes #1334

I've added some logic to snap the right side of the last slide to the right side of the scrolling window.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased](https://github.com/akiran/react-slick/tree/HEAD)
 
-## 0.24.0
+## 0.24.1
 
 **Release Changes**
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## [Unreleased](https://github.com/akiran/react-slick/tree/HEAD)
 
+## 0.24.2
+
+**Release Changes**
+
+- Bug fixes
+  - It removes the need of slidesToscroll/slidesToShow === 1
+  - It fixes the issue of right snapping when the images inside the gallery doesn't fill the sliding window.
+
 ## 0.24.1
 
 **Release Changes**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,18 @@
 
 ## [Unreleased](https://github.com/akiran/react-slick/tree/HEAD)
 
+## 0.24.0
+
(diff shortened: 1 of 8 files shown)


cgadam:
This is nice @diegopamio! Please consider the case when there are few items in the carousel. I can see a bug in this case where the few items (in this case just 2) get aligned to the right instead of left as they are supposed to!

See GIF:
csqm77ty3l

diegopamio:
@cgadam done, now it doesn't snap to the right if the width of all the available images is less than the width of the container.

bgoldfarb:
What is the status of this PR? @diegopamio

pinglinh:
Yeah I agree with @precompiled! This needs to be resolved too :/

ydegtyar:
Will it be merged?

groomain:
Will it be merged one day ? 🙏

diegopamio:
I'm having little to nothing free time for this, If anyone can help on @precompiled request, or if we can do it separately, I can collaborate.

groomain:
I've finally switched to swiper. Thanks !

Charlie91:
@diegopamio, Hi! I faced the same problem that this PR solves, and wouldn't mind to finish your job. Could you add me to collaborators of the fork so I could add commits here?

PR #1349, opened 2018-08-28
Fix playing if autoplay prop is changed

TomOne:
When the `autoplay` prop is changed from `false` to `true`, the autoplay functionality didn’t re-initialize and the slider wasn’t re-rendered.

With this PR, changing the `autoplay` prop from `false` to `true` makes autoplay work again as expected. It doesn’t seem to break anything else. I don’t have a full overview over the code base, so make sure this change doesn’t introduce other issues.


Hari-Shankar-3296:
Its working!!!
check here - <a href="https://codesandbox.io/s/8xk07q082l">https://codesandbox.io/s/8xk07q082l</a>
and this is react-slick version - 0.21.0

And let's give some time for those guys to fix this issue.

PR #1371, opened 2018-09-25
Click events are now working on active slides

Emad-salah:
Click events are now working for slides that are active by adding `visibility: hidden` to the inactive slides and `visibility: visible` to the active ones when they're in fade mode


maddhruv:
hey @Emad-salah can you give me a demo link or regeneration code through which I can check the feature

Issue #1245, opened 2018-05-28
SSR and responsive mode


padupuy:
I already open [this issue on the nextjs repo](https://github.com/zeit/next.js/issues/4398) but it seems to be a problem with react-slick.

**Describe the bug**
When I use next.js with react-slick and responsive options, there is a bug during the rehydration phase, the src attribute of the images are mixed between the slides.

The error message is `Warning: Prop src did not match. Server: "http://via.placeholder.com/350x150?text=7" Client: "http://via.placeholder.com/350x150?text=8"`

**To Reproduce**
I made an example from examples folder : https://github.com/padupuy/next.js/tree/feature/with-react-slick

The code is right here : https://github.com/padupuy/next.js/blob/feature/with-react-slick/examples/with-react-slick/pages/index.js

**Expected behavior**
The image sources must *not* be mixed

**Screenshots**
<img width="1114" alt="capture d ecran 2018-05-16 a 18 29 07" src="https://user-images.githubusercontent.com/715255/40130361-14af5efe-5937-11e8-92ff-507d4bdc9fb3.png">
![capture d ecran 2018-05-16 a 18 31 01](https://user-images.githubusercontent.com/715255/40130439-5294913a-5937-11e8-914e-2532e1927e40.png)

**System information**
 - OS: macOS High Sierra 10.13.4
 - Browser (if applies) : at least chrome and safari
 - Version of Next.js: at least 5.1 and 6.0.2

virajsinha:
I am facing a similar issue with SSR. Were you able to find a fix?

masterGennadij:
Is any solution to prevent this bug? Thank you.

davidsanchez96:
I don't know if is the best approach but I solved this problem by using dynamic import

```
import dynamic from 'next/dynamic'

const SliderComponentWithNoSSR = dynamic(import('../components/SliderComponent'), {
  ssr: false
})

export default () =>
  <div>
    <SliderComponentWithNoSSR />
  </div>
```

AndriiUhryn:
+1

felipetoffolo1:
+1

josheche:
+1

camwes:
+1

goshdarnheck:
+1

chancestrickland:
Following @davidsanchez96's lead, I still wanted the actual content to be rendered on both the server and the client, so this was my approach.

```js
import React, { PureComponent } from 'react';
import dynamic from 'next/dynamic';
[...]

PR #1377, opened 2018-10-05
Added possibility to react on responsive prop changes

rroslaniec:
Right now when we try to render react-slick on SSR with responsive prop it can be mismatch (on SSR react-slick rendering the biggest breakpoint). To fix that responsive prop can be attach to react-slick after mount on client - but react-slick do not react on that changes. This is simple fix for applying breakpoints over time.

### DIFF
--- a/src/slider.js
+++ b/src/slider.js
@@ -32,38 +32,7 @@ export default class Slider extends React.Component {
     //whyDidYouUpdate(React)
     //}
     if (this.props.responsive) {
-      let breakpoints = this.props.responsive.map(
-        breakpt => breakpt.breakpoint
-      );
-      // sort them in increasing order of their numerical value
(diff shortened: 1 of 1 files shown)


maddhruv:
hey @rroslaniec can you please tell me how can I regenerate the use case

chiplay:
@rroslaniec I had the same issue and this worked for me: https://github.com/akiran/react-slick/issues/1245#issuecomment-463250189

rroslaniec:
uoh - that was really long time ago - basically - when I've SSR'ed slider (for example for full-hd screen) would be nice to trigger again calculations (because for example we wanna to display less elements on slider for 'tablet' view). That will cause re-render ofc but still we can recover from that kinda situations.

We at the end embedded slider with dynamic import

PR #1383, opened 2018-10-08
Reset height to auto for vertical carousels when unslick is true

paulathevalley:
When there are fewer slides than `slidesToShow` for vertical carousels, the height of the carousel was ridiculously tall. In this case, the height is being set by `Track` (from `trackProps.trackStyle`), which is normally overridden with an inline style of `height` on `.slick-list` (from `listProps.listStyle`).

See example here: https://codesandbox.io/s/woo4k1k557

I believe this is happening because `unslick` is true, so the `listProps.listStyle` gets blown out completely by this line: `listProps = { className: "slick-list" }`.

This PR: 
- Reset the `height` to `auto` in the `Track` component for vertical carousels when `unslick` is true. 
- Add an example of a vertical carousel with more slides than `slidesToShow`. 
- lint: Add a `semicolon` to the end of imports for consistency.

cc @exogen

### DIFF
--- a/__tests__/testUtils.js
+++ b/__tests__/testUtils.js
@@ -46,11 +46,7 @@ export function createInnerSlider({ noOfSlides, ...settings }) {
   const children = React.Children.toArray(
     createReactSliderChildren(noOfSlides)
   );
-  return (
-    <InnerSlider {...settings} ref={slider => (this.innerSlider = slider)}>
-      {children}
-    </InnerSlider>
-  );
+  return <InnerSlider {...settings}>{children}</InnerSlider>;
 }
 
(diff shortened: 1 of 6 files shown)


__tests__/testUtils.js
```diff
@@ -46,11 +46,7 @@ export function createInnerSlider({ noOfSlides, ...settings }) {
   const children = React.Children.toArray(
     createReactSliderChildren(noOfSlides)
   );
-  return (
-    <InnerSlider {...settings} ref={slider => (this.innerSlider = slider)}>
```
paulathevalley: hmm, this `ref` was causing the test failures with `TypeError: cannot set innerSlider of undefined`

PR #1389, opened 2018-10-13
Test fixes

maddhruv:
- Updated Snapshots
- fixed `LazyLoad` test

### DIFF
--- a/__tests__/testUtils.js
+++ b/__tests__/testUtils.js
@@ -47,7 +47,7 @@ export function createInnerSlider({ noOfSlides, ...settings }) {
     createReactSliderChildren(noOfSlides)
   );
   return (
-    <InnerSlider {...settings} ref={slider => (this.innerSlider = slider)}>
+    <InnerSlider {...settings} ref={React.createRef()}>
       {children}
     </InnerSlider>
   );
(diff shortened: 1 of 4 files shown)


__tests__/testUtils.js
```diff
@@ -47,7 +47,7 @@ export function createInnerSlider({ noOfSlides, ...settings }) {
     createReactSliderChildren(noOfSlides)
   );
   return (
-    <InnerSlider {...settings} ref={slider => (this.innerSlider = slider)}>
+    <InnerSlider {...settings} ref={React.createRef()}>
```
akiran: @maddhruv This might break for older version of react.

Can you update the fix to make it work for both older and newer versions of react

package.json
```diff
@@ -3,16 +3,18 @@
...
+  "files": [
+    "dist",
+    "lib"
+  ],
   "scripts": {
     "start": "gulp server",
     "build": "gulp clean && gulp sass && gulp copy && webpack",
     "prepublish": "babel ./src --out-dir ./lib && gulp dist",
-    "test": "eslint src && jest",
+    "test": "eslint src && jest --updateSnapshot",
```
akiran: @maddhruv We shouldn't update snapshot automatically.

If some tests, we should update snapshots if necessary in interactive mode

akiran: @maddhruv remove --updateSnapshot option

akiran:
Test issues are fixed now. Closing this PR

PR #1424, opened 2018-11-12
Use custom data attributes in custom arrows.

tchamblee:
This PR adjusts the custom props passed to custom next/prev arrows to use custom data attributes.  This prevents warnings like below from being thrown in React 16.

screen shot 2018-11-12 at 11 23 34 am

### DIFF
--- a/__tests__/arrows.js
+++ b/__tests__/arrows.js
@@ -14,8 +14,8 @@ function CustomArrow(props) {
   return (
     <span
       className="sample"
-      data-currentSlide={props.currentSlide}
-      data-slideCount={props.slideCount}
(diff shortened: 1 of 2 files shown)


tchamblee:
@akiran Mind giving this a look over?

maddhruv:
@tchamblee can you please show how does this actually works?

akiran:
@tchamblee you might be using ```{...props}``` in your custom arrows
Ideally you have to destructure the necessary props and pass it to your div
https://github.com/akiran/react-slick/blob/master/examples/CustomArrows.js#L5

We still send props like currentSlide to custom arrows because that info in some usecases to conditionally disable buttons

PR #1445, opened 2018-12-02
Fixed bug prev button not working

misi444:
The slidesToShow variable is not even necessary here, so I removed that too.

I found this bug by creating a carousel with the following props:

variableWidth: true
infinite: false
slidesToScroll: 3
I had 18 slides. When I clicked the next button for the last time, only the last slide showed, which is OK. But from then on the prev button stopped working. I looked up the source and found the bug. Fixed it and the problem went away.

### DIFF
--- a/src/utils/innerSliderUtils.js
+++ b/src/utils/innerSliderUtils.js
@@ -244,7 +244,6 @@ export const changeSlide = (spec, options) => {
   var indexOffset, previousInt, slideOffset, unevenOffset, targetSlide;
   const {
     slidesToScroll,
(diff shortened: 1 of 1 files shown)


misi444:
Well I'm quite sure my changes didn't break the build and cause this error. Let me know if I can be of help.

maddhruv:
@misi444 can you reproduce the bug and share the sandbox link
I can't see the bug as such!

Thanks

yshterev:
@maddhruv the bug is real ... same settings same case.

PR #1449, opened 2018-12-04
updated travis config

maddhruv:


.travis.yml
```diff
@@ -1,3 +1,29 @@
 language: node_js
+
 node_js:
-  - "stable"
\ No newline at end of file
+  - "stable"
+  - "10"
```
akiran: @maddhruv Running tests on node stable will be sufficient.
We needn't run on 3 versions of node

maddhruv: making it safer for about all user (mostly on 6/8/10) and not landing a bad thing (async/await) for someone using an older version
just a precaution

akiran: This is not a nodejs library. Only we run the tests of react-slick.
So we needn't worry about supporting multiple nodejs version.

Stable will be sufficient

maddhruv: Okay gotta
will remove rest

Issue #1331, opened 2018-08-06
customArrow meet a problem


walker-jiang:
**use this:** 
    var settings = {
      dots: false,
      infinite: true,
      speed: 500,
      slidesToShow: 6,
      slidesToScroll: 1,
      nextArrow: <DirectIcon type="right" />,
      prevArrow: <DirectIcon type="left" />
    };
**get this:** 
   Warning: React does not recognize the `currentSlide` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `currentslide` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
   React does not recognize the `slideCount` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `slidecount` instead. If you accidentally passed it from a parent component, remove it from the DOM element.

That's not a big problem, but hope to solve.thanks

paulcredmond:
It would be more helpful to have a JSFiddle or CodePen to demo your issue.

Here's an example of using a custom arrow where you have access to `currentSlide `and `slideCount`.

In your settings:

`nextArrow: <NextArrow />,`

Your arrow function:

```
// Next arrow
function NextArrow(props) {
   const {currentSlide, slideCount, className, onClick} = props;

   return (
      <div onClick={onClick}>
         <button className={"slick-arrow slick-next"} />
      </div>
   );
```

alyrik:
@walker-jiang, only components are allowed to be a custom arrow. You can't provide plain JSX tags to `prevArrow` and `nextArrow` props.

Issue #1238, opened 2018-05-22
CurrentSlide and slideCount Warnings still in v 0.23.1


waheed25:
I am still getting these warnings i am passing arrow like this
`prevArrow: LeftArrow(), `
LeftArrow is a svg icon.

yoyo837:
Yes，it's work. https://github.com/akiran/react-slick/issues/728#issuecomment-310773471

AgainPsychoX:
I have the same issue I guess?

AgainPsychoX:
In DOM tree (Chrome DevTools):
![image](https://user-images.githubusercontent.com/23557839/124244389-2e2b7080-db1f-11eb-87d0-3129d5f9db6b.png)

Errors (Chrome console):
![image](https://user-images.githubusercontent.com/23557839/124244460-40a5aa00-db1f-11eb-8e4b-3e85d6615b2c.png)


Used settings for Slider:
```
				infinite 
				speed={500}
				slidesToShow={3}
				slidesToScroll={1}
				responsive={[
					{breakpoint: 1200, settings: {
						slidesToShow: 2,
					}},
					{breakpoint: 940, settings: {
						slidesToShow: 1,
					}},
				]}
				arrows
				nextArrow={
					<div>
						<NavigateNext fontSize="inherit"/>
					</div>
				}
				prevArrow={
					<div>
						<NavigateBefore fontSize="inherit"/>
					</div>
				}
```

The `NavigateBefore`/`NavigateNext` are SVGs, nevertheless I packed those in `div`s since I got some other problems (see issue #2051 )

Issue #1195, opened 2018-04-18
Warning: React does not recognize the `slideCount` prop on a DOM element.


kirill-konshin:
Because of this line https://github.com/akiran/react-slick/blob/master/src/arrows.js#L35 I get errors with React 16.

![image](https://user-images.githubusercontent.com/1530410/38920602-7bd50f66-42a8-11e8-9b77-0e2a6152d4b1.png)

What is the purpose of this? How it will help with a config like this:

```
    nextArrow: (
        <span>
            <IconButton>
                <FA icon={faChevronRight} />
            </IconButton>
        </span>
    ),
```

laveesingh:
Don't pass down all the props that you receive for custom arrows. 
```js
nextArrow: ({ firstProp, secondProps, ...otherProps }) => (
  <element {...props that you want} />
)
```
Something like this should work.

kirill-konshin:
Any progress?

Pushplaybang:
would be great to solve this.

kirill-konshin:
You may use this as workaround:

```js
const SlickButtonFix = ({currentSlide, children, ...props}) => <span>{children}</span>;

const SETTINGS = {
    prevArrow: (
        <SlickButtonFix>
            <IconButton>
                <ChevronLeft />
            </IconButton>
        </SlickButtonFix>
    )
};
```

akashmdkml:
@kirill-konshin  thanks this worked well  👏, react-slick custom arrows help to make arrow accessibility. https://codesandbox.io/s/myqq0yz54y

bohehe:
> ```js
> <SlickButtonFix>
>             <IconButton>
>                 <ChevronLeft />
>             </IconButton>
>         </SlickButtonFix>
> ```

Thank you, man! you saved my life!

codfish:
@bohehe @kirill-konshin I'm running into seeing the button appear along my custom arrow with this fix, did you run into this at all?

![image](https://user-images.githubusercontent.com/1666298/102720674-b5bc3b00-42c3-11eb-905d-f6ec23462686.png)


```jsx
const SlickButtonFix = ({ currentSlide, slideCount, children, ...props }) => (
  <span {...props}>{children}</span>
);
```

```jsx
prevArrow={
  <SlickButtonFix>
    <img src="/chevron-left.svg" alt="previous quote" width="12" height="20" />
  </SlickButtonFix>
}
nextArrow={
  <SlickButtonFix>
[...]

Issue #1130, opened 2018-03-22
Custom Next and Prev Props logs warnings to the console.


Pushplaybang:
The following two errors occur when using custom elements for the next and prev arrows.

```
Warning: React does not recognize the `currentSlide` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `currentslide` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```

```
warning.js:33 Warning: React does not recognize the `slideCount` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `slidecount` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```

I've defined them as stateless functions as follows:

```js
export const PrevArrow = props => (
  <button type="button" {...props}>
    <PrevIcon />
  </button>
);

export const NextArrow = props => (
  <button type="button" {...props}>
    <NextIcon />
  </button>
);
```

laveesingh:
First of all, these are warnings, not errors.
The props that are passed to arrow components, contain slideCount and currentSlide custom props. So, if you're gonna define your arrows like this, you might wanna extract those props out first. Something like this should do:
```javascript
const {currentSlide, slideCount, ...arrowProps} = props
<button {...otherProps}>
```

Pushplaybang:
@laveesingh  - thanks for the answer!

did you mean:

```js
const {currentSlide, slideCount, ...arrowProps} = props
<button {...arrowProps}>
```

did note them as warnings in the title, and have corrected it in the body.

minbelapps:
Why is it closed?  This error is fired when JSX elements are provided instead of functions:

```
PrevArrow :  (
  <button type="button">
    <PrevIcon />
  </button>
);
```

PR #1453, opened 2018-12-06
Allow custom arrow to be React element

cheeZery:
Former to this commit using a React element directly as an prev or next arrow resulted in React errors due to props unknown to the React elements they were applied to.
The common props (like className and onClick) are still attached to the cloned React element, but won't be overwritten, if they are already defined.

Given the amount of people trying to use simple React elements as custom arrows shows the need for this fix. This PR will solve the following issues (which were mostly just closed without a real solution):

- #623
- #728
- #780
- #1130
- #1195
- #1238
- #1331

### DIFF
--- a/__tests__/arrows.js
+++ b/__tests__/arrows.js
@@ -4,7 +4,7 @@
 
 sinon.stub(console, "error");
 
-import { render, shallow } from "enzyme";
+import { render, mount, shallow } from "enzyme";
 import React from "react";
 import sinon from "sinon";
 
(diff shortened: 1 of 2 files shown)


__tests__/arrows.js
```diff
@@ -32,6 +32,34 @@ describe("Previous arrows", () => {
...
+      );
+    };
+
+    const wrapper = mount(<PrevArrow prevArrow={<CustomArrow />} />);
+
+    expect(wrapper.find("span.customArrow")).toHaveLength(1);
+    expect(console.error.called).toBe(false);
+  });
+
+  it("sould render React element custom arrow", () => {
```
ahahn95: typo

PR #1456, opened 2018-12-07
fix surplus closure

alex3347:
Variable prevClickHandler can't be released because of closure. And function checkImagesLoad will execute at least 8 times each movement in autoplay mode. In my project, it will add 200kb usage of memory in website each movement and my project is used in webview on iOS platform.  After monitoring usage of memory in iPhone，we found 200kb usage of memory in website can cause 2M usage of memory in webview in iPhone.  And it will cause crash of webview after moving some times.


nchu-rb:
bumping this request for an approval as well. thanks.

Issue #1172, opened 2018-04-05
Responsive feature using unslick wipes out containing divs


yenly:
Example code here:
https://codesandbox.io/s/j1l1nn0695
https://z613xn9pk3.codesandbox.io/

Please look at the above example and change the browser size to see this critical bug.

Version 22.3 didn't have this issue. Latest version replaces slide content with empty div. 
<img width="1550" alt="screen shot 2018-04-05 at 11 29 40 am" src="https://user-images.githubusercontent.com/6759658/38384878-286952fc-38c5-11e8-8433-b1e7aa6a050d.png">
As you can see in the above screen capture, once the browser is in the setting for responsive unslick, it hits the  zone where everything in <div class='slick-slider> becomes an empty div.

danilobuerger:
Can confirm, this worked for me in 0.22.3 and is broken afterwards.

misstricky:
Any word as to when this fix is going to be released? Thanks!

ryanboswell:
I don't usually like to "bump" things, but any hope for a release with this fix sometime soon?

johnrom:
I can confirm that the fix in 56b124da9509370a33682677f205c2092f59b648 works

AndyArcherKG:
Is there a fix of this live now? 
i'm checking `0.23.1` and when i "unslick" it wipes all my content.

misstricky:
Yes I think unslick is fixed, but, to work around what I think you're experiencing, I just went ahead and skipped all unslick behavior and put resizes listeners in my component, and then only initiate the slideshow when I need it, something like...

```
 getSettings() {
        return {
            slidesToShow: 1,
            slidesToScroll: 1,
            centerMode: true,
            centerPadding: this.props.centerPadding,
            speed: 300,
            arrows: false,
            adaptiveHeight: false
        };
    }

    render() {
        const settings = this.getSettings();
        let filterNull = React.Children.toArray(this.props.children).filter(
            (item) => item !== null
        );

        if (
            this.state.windowWidth <= this.props.startBreakpoint &&
            filterNull &&
            filterNull.length > 1
        ) {
            return <Slider {...settings}>{this.props.children}</Slider>;
        } else {
            return <div>{this.props.children}</div>;
        }
    }
}
```

This is probably not the most helpful example, but, the react-slick `Slider` will only render if the window is less than the breakpoint prop here, otherwise do not render slideshow. Just not using unslick anymore.

AndyArcherKG:
[...]

PR #1458, opened 2018-12-08
Fix a bug that `settings: "unslick"` renders nothing #1172

mskims:
In case of `settings: "unslick"`, newChildren will be empty list. because of these lines:
https://github.com/akiran/react-slick/blob/48c93ed36cfb58991e6bf4d34a1a588094281d12/src/slider.js#L94-L97
https://github.com/akiran/react-slick/blob/48c93ed36cfb58991e6bf4d34a1a588094281d12/src/slider.js#L163-L167 

So make slider returns natural `children`, instead of `newChildren`

p.s. As i commented on https://github.com/akiran/react-slick/issues/1172#issuecomment-445483592, https://github.com/akiran/react-slick/commit/b85f15672548a78cd1407331b068615711edd98f#diff-e8eaa42a1b8ab4981d9272be425c03dfL121 may broke this function.

### DIFF
--- a/src/slider.js
+++ b/src/slider.js
@@ -203,7 +203,7 @@ export default class Slider extends React.Component {
 
     if (settings === "unslick") {
       const className = "regular slider " + (this.props.className || "");
-      return <div className={className}>{newChildren}</div>;
+      return <div className={className}>{children}</div>;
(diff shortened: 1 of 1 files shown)


mskims:
Found that issues had already fixed at dev branch. Closing.

PR #1465, opened 2018-12-12
Fixbug arrows.js bug

gyh9457:
1. currentSlide and SlideCount Warnings. #1238
2. customArrow can not get the class and style

### DIFF
--- a/src/arrows.js
+++ b/src/arrows.js
@@ -32,8 +32,8 @@ export class PrevArrow extends React.PureComponent {
       onClick: prevHandler
     };
     let customProps = {
-      currentSlide: this.props.currentSlide,
-      slideCount: this.props.slideCount
+      currentslide: this.props.currentSlide,
+      slidecount: this.props.slideCount
     };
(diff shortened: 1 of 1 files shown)


gyh9457:
> It's fine, but you can try create variable prevArrow and nextArrow from destructuring before creating variables with styles and className. e.g. `const { prevArrow } = this.props` Thanks to this, the statement will be simpler. e.g. `const prevArrowStyle = (prevArrow && prevArrow.props && prevArrow.props.style) || {};`

Thank you for your advice, if we want to make the statement simpler, I think we should refactored the `render` function. e.g. 
```
    const { prevArrow, infinite, currentSlide, slideCount, slidesToShow } = this.props;
    ......
    let prevArrowElement
    ......
```
Maybe I'll create a new PR after this PR was merged.

mczeplowski:
That would be the best.

Lelith:
Is there any update on this PR? Because i still get the error.

Issue #1160, opened 2018-03-30
feature request: capture & prevent click events from reaching elements within slides, if user swipes


nihlton:
When the entire slide content is clickable, it becomes impossible to swipe between slides without also triggering the click.

it would be nice if the slider would swallow the click event when it detects the user is swiping between slides.

laveesingh:
It has been implemented recently. The changes should be released in a day or two.

akiran:
Released 0.23.0

Tuizi:
Is there any doc about that? Or a commit version? And when the version 0.23 will be release?

JeffWeim:
@akiran 

Was this fix ever released? it looks like the latest version is only at 0.21.0 (https://github.com/akiran/react-slick/tags). Is this coming in a future release?

Thanks

akyker20:
I don't think this was ever fixed. I have 0.23.1 and when I swipe, onClick of my slide content is triggered.

arkolesen:
@akiran @laveesingh 
Hi, guys!
You component is very cool! thanks for it!
Could you help us (@JeffWeim, me, and @akyker20)?
Thanks a lot!

arkolesen:
@akyker20 Hi!
Have you dealt with it?

all9lives:
@akiran Did this fix ever get merged in? I looked and couldn't see the changes in `0.23.1`, it doesn't look like there's a `0.23.0` release.

arkolesen:
@all9lives Seems - nope ._.

arkolesen:
@all9lives I dealt with this by:
Adding   `toggleDragging={this.toggleDragging}` property to Carousel component.
This.toggleDragging:

```
  toggleDragging = () => this.setState({isDragging: !this.state.isDragging});
```

and pass flag `isDragging={this.state.isDragging}` to each element of carousel.

Then add onClickHandler to carousel element:

```
 onClickHandler = (event) => {
    if (this.props.isDragging) {
      event.preventDefault();
    } else {
      ... your stuff...
    }
  };
```
and call it as
```
      <div className="blah blah" onClick={this.onClickHandler}>
```

Hope it will help you.

Issue #848, opened 2017-09-04
Drag on sliders fires a click event


GiancarlosIO:
When you drag a slider with a tag "<a href="www.google.com />" slick fires a click to finish the drag.
How can i prevent that?

example:
https://jsfiddle.net/nexuszgt/pgoap3bf/4/

thanks.

costagolub:
You can achieve the desired effect by using `beforeChangeand ` `afterChangemethods `and assign an `onClick `on the anchors.
Based on your sample code this should be working and will not force links to open on dragging. You can also use ES6 classes instead and bind to `this `instead of a variable.
https://jsfiddle.net/eckrawhy/

GiancarlosIO:
@costagolub your "hack" works but i think slick-lib should be fix that because it's a behavior we don't expect

boltcoder:
has this been fixed?

iamgreyy:
> has this been fixed?

Its not fixed on my 0.23.1 version

VeronQ:
I had the same problem, which I solved for my case with the following code.

```javascript
$('.my-slider').on('swipe', function(event) {
  $(this).addClass('on-drag');
});

$('.my-slider').on('afterChange', function(event) {
  $(this).removeClass('on-drag');
});

$('.my-link-slide').click(function(e) {
  if ( $('.my-slider').hasClass('on-drag')) {
    e.preventDefault();
  }
});
```

akashmdkml:
> Here's my solution that makes use of clientX and clientY attributes of the `click` and `mousedown` events and react component state to solve this issue.
> 
> Basically it's comparing the mouse position at mouse down and click and if either X or Y is different, then it will `preventDefault()` (default browser behavior being to go to the link)
> 
> ```
> class Carousel extends React.Component {
>   constructor (props) {
>     super(props)
>     this.state = {
>       clientXonMouseDown: null,
>       clientYonMouseDown: null
>     }
>     this.settings = {
>       dots: true,
>       arrows: false,
>       swipeToSlide: true,
>       infinite: false,
>     ...
>     }
>   }
> 
>   handleOnMouseDown (e) {
>     this.setState({
>       clientXonMouseDown: e.clientX,
>       clientYonMouseDown: e.clientY
>     })
>     e.preventDefault() // stops weird link dragging effect
>   }
> 
>   handleOnClick (e) {
>     e.stopPropagation()
>     if (this.state.clientXonMouseDown !== e.clientX || 
[...]

Issue #604, opened 2017-01-19
Slider content should not be clickable while the user is dragging the slides on Desktop.


fdc-viktor-luft:
I replicated the issue here: https://jsfiddle.net/20bumb4g/1988/

All pictures were wrapped into a link which "would" link to an external page. (google at this example)

If you start dragging any slide without changing the slide after releasing the click (no index update!) the event will be propagated down to the inner content. This makes dragging act differently and unwanted on desktop.

Please fix this issue. Btw.: your library is awesome.

minooo:
@akiran Please solve this problem

adamshone:
Maybe you could pass the event through to the `beforeChange` callback so that we can call `stopPropagation` on it?

minooo:
@adamshone  Good idea, but I found that still doesn't work as the `beforeChange` accepts only two parameters  `currentSlide` and `nextSlide`.

adamshone:
Sorry, by "you" I was referring to the project authors, I thought they could add the event as a third param to `beforeChange`.

Having had a very quick scan of the code I think unfortunately it's not that simple, it looks like it's using some `animationEnd` callbacks rather than listening to any mouse event that could be passed on to the user :/

adamshone:
I tried using the `beforeChange` and `afterChange` event handlers to toggle a flag `this.setState({ignoreClicks: true/false})`, and pass `state.ignoreClicks` to the carousel items as a prop.

Unfortunately triggering a render from the `beforeChange` event seems to break the carousel, it cuts the animation off and seems to reset some internal state so that `afterChange` is never fired.

So... I have a workaround but it's *very* hacky - you can use it if you're desperate:

```
import Carousel from 'react-slick';

class MyCarousel extends PureComponent {
  midDrag = false;

  toggleMidDrag = () => {
    this.midDrag = !this.midDrag;
  }

  render() {
    return (
      <Carousel beforeChange={this.toggleMidDrag} afterChange={this.toggleMidDrag}>
        <Item key="0" parentCarousel={this} />
        <Item key="1" parentCarousel={this} />
        ...
      />
    );
}

class Item extends PureComponent {
  onClick = () => {
    if (this.props.parentCarousel.midDrag) {
      // ignore clicks
      return;
    }

    // ...your usual click handling code
  }

  render() { ... }
}
```

erasmo-marin:
@akiran please, this is critical. I think the easier solution is to stop the event propagation on dragEnd in the event handler mixin.

[...]

Issue #542, opened 2016-11-11
Slide as link vs mouseDrag


vdv73rus:
So, my slides are links (`<a>` tag).

And react-slick does not understand whether our click is 'click on the link' or 'click as dragstart'.
So, browser go to other page when I start drag the slides.

102:
@vdv73rus have you solved this?

Deepblue129:
+1

superfunkminister:
+1

cmalonzo:
+1

minooo:
@akiran please solve this problem

ananevam:
+1

edwardfhsiao:
+1

KROT47:
+1

DavidKk:
+1

CruelSilence:
+1

processprocess:
+1

edwardfhsiao:
I found a solution might help.

My scenario is that I want the image to be clicked and do something, but when I drag it to next slide, it would trigger the event anyway. In order to specify the click and drag, change the code as below
```js
            <img
              ref={ref => {
                this.img[k] = ref;
              }}
              src={require('some.jpg')}
              onMouseDown={() => {
                this.img[k].eventType = 0;
                console.log(`onMouseDown ${this.img[k].eventType}`);
              }}
              onMouseMove={() => {
                this.img[k].eventType = 1;
                console.log(`onMouseMove ${this.img[k].eventType}`);
              }}
              onMouseUp={() => {
                console.log(`onMouseUp ${this.img[k].eventType}`);
                if (this.img[k].eventType === 0) {
                  alert(1);
                  console.log('click');
                } else if (this.img[k].eventType === 1) {
                  console.log('drag');
                }
                this.img[k].eventType = 0;
              }}
            />
```

laveesingh:
Commit bc8a5be should solve the problem.
Changes will be released soon.

n0umankhan:
not resolved yet ?

joebentaylor1995:
Still doesnt work for me

PR #1483, opened 2019-01-13
Stop child component's onClick event when Slider is dragged.

lonyele:
Hi, I recently have been using this library, it worked fine for my use cases except this behaviour of triggering child component's onClick event after Slide is dragged or swiped. Similar problems that have been issued here are #848, #1160, #542, #604    

User of the library can manually block onClick event by making a dragging state on their own way using beforeChange(), afterChange(), I think this can be implemented in library by default, because I found this is a common feature that many people would want it right away using this library

My solution is so simple that It adds one state called `previouslyDragged` boolean and 
on this.swipeStart(), sets `previouslyDragged` to false
on this.swipeMove(), sets `previouslyDragged` to true
on onClickCapture() event, stop propagation if previous state was dragged or swiped

Now, just clicking each item of slide works as expected and after the slider is dragged or swiped It won't trigger child component's onClick event

### DIFF
--- a/src/inner-slider.js
+++ b/src/inner-slider.js
@@ -36,7 +36,8 @@ export class InnerSlider extends React.Component {
     this.state = {
       ...initialState,
       currentSlide: this.props.initialSlide,
-      slideCount: React.Children.count(this.props.children)
+      slideCount: React.Children.count(this.props.children),
+      previouslyDragged: false
     };
     this.callbackTimers = [];
     this.clickable = true;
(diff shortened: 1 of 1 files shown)


rafbc:
Any ETA to when this will be merged?! :P

young666:
This doesn't work when I use <a> instead of <div> in Slider, I have try disable drag and drop for <a>, still doesn't work 😟

rafbc:
I used this
```
constructor(props) {
		super(props);

		this.state = {
			previouslyDragged: false
		}
	}

	handleImageClick(e) {
		const { previouslyDragged } = this.state;
		if (previouslyDragged) {
			e.stopPropagation();
			return;
		} else {
			console.log('CLICK FINALLY WORKS')
		}

	}
```


```

		const settings = {
			dots: true,
			infinite: true,
			speed: 500,
			arrows: true,
			slidesToShow: 1,
			slidesToScroll: 1,
			responsive: [
				{
					breakpoint: 768,
					settings: {
						slidesToShow: 1
					}
				},
				{
				breakpoint: 480,
					settings: {
						slidesToShow: 1
					}
				}
			],
			beforeChange: () => {
				this.setState({ previouslyDragged: true })
			},
			afterChange: () => {
				this.setState({ previouslyDragged: false })
			}
		};

<Carousel {...settings}>
....
</Carousel>
```

young666:
@rafbc it only works if after drag I move my mouse away from the item I dragged 😃

lonyele:
@young666 You are right. I can fix <a> tag issue by adding "event.preventDefault()" to where I added "event.stopPropagation()".  What I need was to block events thus I put only stopPropagation, but If there is a need to stop browser's default actions then preventDefault() is also necessary.

After I added preventDefault(), I found <a> tag's link doesn't work when dragged. However It is really janky and still I can drag text of <a> tag's link.  After I set the draggable={false} on <a> tag, text is not draggable.

I added an example at the bottom so that you can check the behaviour of each cases

akashmdkml:
@lonyele  Hey nice work, I tried your fix, now it does not fire the click event when I drag or swipe. but now the drag feature is hard to drag experience, especially when I swipe from left to right it does not swipe properly. let me know if you have any suggestion for this. Thanks. 
here is my demo code https://codesandbox.io/s/myqq0yz54y

deleterepo:
Any updates on this one? I see the changes were approved but not merged yet.

lonyele:
Sadly, it seems like this project is not maintained anymore(nothings get merged recently)

raphaelbs:
Any chance this will get merged?

lonyele:
@akiran just a gentle ping...

Issue #1484, opened 2019-01-15
[bug] slider breaks on SSR when vertical is true


tompere:
### Issue Reproduction
https://codesandbox.io/s/9wrmql2xo
(or see [live site](https://9wrmql2xo.sse.codesandbox.io/))

### Description
- slider **breaks** on SSR + `vertical: true`.
- SSR works **as expecte**d when  `vertical: false`

## React SSR render (using next.js)
![screen shot 2019-01-15 at 16 34 59](https://user-images.githubusercontent.com/2641321/51187059-acf6a980-18e3-11e9-99ed-88e9840ffad1.png)

## React Browser render
![screen shot 2019-01-15 at 16 35 17](https://user-images.githubusercontent.com/2641321/51187060-acf6a980-18e3-11e9-915c-514c255e8afc.png)

## slick settings
```
{
    dots: true,
    infinite: false,
    vertical: true,
    verticalSwiping: true,
    slidesToShow: 3,
    slidesToScroll: 1
  }
```

romanown:
I have SSR and the slider is work on vertical. https://codesandbox.io/s/9j7x86x8mr

sidzan:
I also encountered the same issue, were you able to fix it or find any alternative solution?

Orient87:
And I have the same problem, maybe someone has a solution?

gorork:
Same issue. Any solutions?

PR #1490, opened 2019-01-24
Fix vertical ssr flow

alex5il:


alex5il:
Fixes [#1484](https://github.com/akiran/react-slick/issues/1484)

PR #1493, opened 2019-01-28
Persist custom arrow className

pnarielwala:
Hello 👋 

I noticed in my project that the class I have on my arrow component disappears when trying to use custom components (I'm using css modules)
```js
const settings = {
 nextArrow: (
    <div className={styles.arrowContainer}>
        <ArrowRight className={styles.arrow} />
    </div>
  )
}
```
I would like to remove the `::before` styling on `.slick-prev` and `.slick-next` (from [slick-carousel](https://github.com/kenwheeler/slick/blob/master/slick/slick-theme.css#L89)), but I can't 😔

So, I opened this PR in hopes to fix that. Other than that, thanks for this amazing library!

Parth

### DIFF
--- a/src/arrows.js
+++ b/src/arrows.js
@@ -40,7 +40,8 @@ export class PrevArrow extends React.PureComponent {
     if (this.props.prevArrow) {
       prevArrow = React.cloneElement(this.props.prevArrow, {
         ...prevArrowProps,
-        ...customProps
+        ...customProps,
+        className: classnames(prevArrowProps.className, this.props.prevArrow.className)
       });
     } else {
       prevArrow = (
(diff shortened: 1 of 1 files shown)


pnarielwala:
https://github.com/akiran/react-slick/pull/1465
☝️I guess this fixes that too

chemeseddine-digit:
Oh wow, it was one year ago ? Is this project still able to have some improvements or should we think about switching to another lib instead or help reviewing ?

It will be a pleasure to help you guys :+1: 
If this repo is closed, please then notify the community :smiley_cat:

pnarielwala:
@akiran is this something we can add or should I close this out?

PR #1495, opened 2019-01-30
Lazyload closest images

wendywy0660:
Hello there! 

I am using `react-slick` for our image sliders on the search page. Since there are 20 sliders and we use `lazyload: true`, it gave a white flash when you click next/prev ( a right to left transition to an empty rectangle before the next image loads) 

So I just created this PR to add option to enable lazy load with preloading the closest slides images (for previous and next slide images). Since the `lazyload` prop already takes string `progressive`, I overload it by passing a number (the number of images you would like to preload). Ideally it could take an object to config.

Could you please review and let me know your thought?

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased](https://github.com/akiran/react-slick/tree/HEAD)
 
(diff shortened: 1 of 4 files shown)


albertstill:
+1 to this!

PR #1496, opened 2019-02-04
Lazyload closest images using array

wendywy0660:
Hello there!

I am using react-slick for our image sliders on the search page. Since there are 20 sliders and we use lazyload: true, it gave a white flash when you click next/prev ( a right to left transition to an empty rectangle before the next image loads)

So I just created this PR to add option to enable lazy load with preloading the closest slides images (for previous and next slide images). Since the lazyload prop already takes string progressive, I overload it by passing an array with two integers, first one is number of items you would like to preload before current image, and second one is after the current image. Ideally it could take an object to config.

Could you please review and let me know your thought?

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 **Release Changes**
 
 - Internal Changes
+  - added option to enable lazy loading for the closest slides images (for previous and next slide images)
   - converted InnerSlider from createReactClass object to ES6 class
   - removed all the mixins, created classMethods and pure utility functions instead
   - changed autoplay from setTimeout to setInterval
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,17 @@
 # Change Log
 
 ## [Unreleased](https://github.com/akiran/react-slick/tree/HEAD)
+- Note
+  - Project forked by Domain at this point to maintain this codebase.
+
+- Added 
+  - option to enable lazy loading for the closest slides images (for previous and next slide images) See example `LazyLoadClosestImages`
 
 ## 0.22.0
 
 **Release Changes**
 
 - Internal Changes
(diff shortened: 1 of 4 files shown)


sergiubologa:
@akiran can you please review this change? Seems like a very small change and shouldn't take too long to review it, and the benefit is huge in terms of performance.

sergiubologa:
@akiran any updates on this?

aks1994:
This is great, can this be merged?

flieks:
anyone can review it ? ;)

jooj123:
@akiran Can this one be merged please?

hay12396:
Anything on this yet?

akiran:
@wendywy0660 We want to maintain API compatibility with slick-carousel library
https://kenwheeler.github.io/slick/

We don't add new props which are not support by it

akiran:
Looks like our lazyLoad functionality is different from the slick-carousel.
lazyLoad option in it accept 'ondemand' and 'progressive'

If lazyLoad functionality in slick-carousel works for you, please a send PR which mimic the same behaviour

pesimeao:
> Looks like our lazyLoad functionality is different from the slick-carousel.
> lazyLoad option in it accept 'ondemand' and 'progressive'

It also supports the `anticipated` value, that loads the next slide only (it's not in the documentation, but it exists)

hayzohargettacar:
hey @pesimeao i tried putting the `anticipated` value but it doesnt seem to work..

flieks:
@hayzohargettacar this should work. 
check https://github.com/kenwheeler/slick/issues/3393 or http://jsfiddle.net/dyfpt14j/
and react-slick is using the latest slick.

I'm using ant design carousel which uses react-slick, it doesn't work their so i want to be sure it's not a antd issue

PR #1516, opened 2019-03-15
LazyLoad preloading next/previous slides by configuration

idangozlan:
Now LazyLoad is available with preloading next/prev slides as configured.

All you have to do is set the following props:
```
      lazyLoad: "offset",
      lazyLoadOffset: 1,
```

lazyLoadOffset will configure the number of next/prev slides to preload.


pesimeao:
Any reason why this was rejected?

flieks:
Also would want to know..

PR #1524, opened 2019-03-28
Add activeDotClass to props

DannyAziz:
The activeDotClass prop accepts a string that will be used instead of `slick-active` for the active dot

It defaults to `slick-active` if nothing is passed to `activeDotClass`


akiran:
Can't add new settings that are not present in [slick](https://kenwheeler.github.io/slick/)

PR #1543, opened 2019-04-26
Stick to right for 'finite + variableWidth' mode.

und3rdg:
Resolve problem with this stupid extra blank space on the right when you going to last slide.
There was no solution for responsively stop last slide sticked to right.

There are some workaround for it, but no for variableWidth scenarios.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,6 @@ jquery.html
 docs/fonts/
 docs/ajax-loader.gif
 package-lock.json
+*.vim
+*.swo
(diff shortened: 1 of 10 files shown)


emregurler:
yes, it works just fine when variableWidth=false. It wraps without any space on the right. It should be fixed for variableWidth. 
For example: Lets assume a page of slider can contain 4 components and there are 17 components. On last page, it shows just one component on left, and breaks 3 component spaces on the right.

Issue #1221, opened 2018-05-10
Prevent user from advancing slide until condition is met


paulcredmond:
Not a bug, just a question on functionality.

I want to use the behaviour you get when swiping the last slide in the slideshow, but not just at the end, but on each slide unless a certain condition is met.

E.g. You have a slide with a form field, when the field is not valid and you try and swipe, you can see some of the next slide but it won't let you advance and bounces back, like the last slide of a normal slider. When the form field is valid, you can swipe the slide as normal.

Anyone found a solution to be able to do this?

`onSwipe` fires before `beforeChange` so I can't get the current slide index - which I could use to check if that slide is valid. I could use `afterChange` to move the slider back one slide, but that would be allowing the user to advance to the next slide, which I don't want.

Any advice appreciated!

mcfoton:
what about hiding the arrows and calling **slickNext()** after validation? 
https://react-slick.neostack.com/docs/api/#methods

paulcredmond:
@mcfoton Yep that's what we're doing at the moment, but the thing is we're also allowing swipe. We want the user to be able to review each slide by sliding back, but if they change a field they cannot swipe left until the field passes validation again. The only option for swipe is true or false, I'd love to be able to stop the user swiping to the left or let the swipe left animation almost complete but they bounce back.

Dobby89:
I'd also like this functionality. I tried using `beforeChange` to prevent the slide from advancing, but it doesn't work but it didn't work.

@paulcredmond Did you manage to come up with a solution (short of dynamically adding a slide)?

paulcredmond:
@Dobby89 We did but it's so long ago I can't remember and don't have access to the code, sorry :(

PR #1548, opened 2019-05-06
Prevent slide change when beforeChange return false

achorein:
Useful for custom behaviors : 
- #1221 : Prevent user from advancing slide until condition is met
- ignoring some slide when "inactive"

Example :
```javascript
beforeChange(oldIndex, newIndex) {
   if (newIndex === 5) {
     this.slider.slickGoTo(6);
     return false;
   }
 }
```

### DIFF
--- a/src/inner-slider.js
+++ b/src/inner-slider.js
@@ -384,7 +384,12 @@ export class InnerSlider extends React.Component {
       useCSS: this.props.useCSS && !dontAnimate
     });
     if (!state) return;
-    beforeChange && beforeChange(currentSlide, state.currentSlide);
+    if (beforeChange) {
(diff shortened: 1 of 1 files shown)


akiran:
beforeChange is just event callback to trigger before slide change
We can't use it to stop slide change

This will break api compatibility with slick-carousal
https://kenwheeler.github.io/slick/
So, we can't accept this PR

Badbreaddead:
@akiran how to stop slider then? Looks like there is no control on it from any api...

Dobby89:
@akiran Can we get a method which prevents the user from going to a slide please? It would be really useful.

andrii-shpontak:
I, too, have spent too much time on this problem and still haven't found a solution...

PR #1555, opened 2019-05-15
Add slide offset value to swipeEvent callback

scailbc:
I pass a `slideOffset` value to the swipeEvent callback. The value is the difference between the previous selected slide and the slide that, due to a swipe, currently appears as the current slide.

```
swipeEvent = ( swipeDirection, slideOffset ) => {
    console.log("Direction  "+swipeDirection+" with offset "+slideOffset);
    // Direction right with offset -2
}
```

This allows to animate a slide during a swipe. The swipe is a translate of the whole carousel, so the current slide is not updated until the swipe is over.

I've also created an example, see [/examples/SwipeAnimated.js](https://github.com/scailbc/react-slick/blob/13c38c80fea70e64a7ef10470aab011fb32cebb8/examples/SwipeAnimated.js)


scailbc:
Hello @akiran, I would like to see this PR merged to master. Why it has not be merged already? There is something that I can do?
Thanks

PR #1561, opened 2019-05-23
updated var(s) to let and const 

Hillsie:
to follow popular view that var should be deprecated for the following reasons: read these posts https://wesbos.com/let-vs-const/ and  https://mathiasbynens.be/notes/es6-const


Hillsie:
Hi @akiran . Updated vars to let and const. Also added some arrow functions instead of function syntax.

PR #1581, opened 2019-06-27
Move onclick and onload methods on component level, removed checkimag…

AndreiSDumitrescu:
…esload from componentDidUpdate


AndreiSDumitrescu:
>

PR #1604, opened 2019-07-30
Update webpack dist config to support umd again

### LINKED ISSUE #1603: Webpack issue from version 0.23 to 0.24
### Guidelines for posting a new issue
On version `0.23`

     git checout 0.23.2
     npm run prepublish

The output in `dist/react-slic.js` will start with

```
(function webpackUniversalModuleDefinition(root, factory) {
        if(typeof exports === 'object' && typeof module === 'object')
                module.exports = factory(require("react"), require("react-dom"));
        else if(typeof define === 'function' && define.amd)
                define(["react", "react-dom"], factory);
        else if(typeof exports === 'object')
                exports["Slider"] = factory(require("react"), require("react-dom"));
        else
                root["Slider"] = factory(root["React"], root["ReactDOM"]);
})(this, function(__WEBPACK_EXTERNAL_MODULE_2__, __WEBPACK_EXTERNAL_MODULE_4__) 
      ^
      |
      NOTE: in this version we had this which is ok considering libraryTarget: "umd",
```

But now with the webpack we got:
```
     deleted:    src/less/atoms/reset.less                                                      │npm ERR!     <https://npm.community>
(function webpackUniversalModuleDefinition(root, factory) {
        if(typeof exports === 'object' && typeof module === 'object')
                module.exports = factory(require("react"), require("react-dom"));
        else if(typeof define === 'function' && define.amd)
                define(["react", "react-dom"], factory);
        else if(typeof exports === 'object')
                exports["Slider"] = factory(require("react"), require("react-dom"));
        else
                root["Slider"] = factory(root["React"], root["ReactDOM"]);
})(window, function(__WEBPACK_EXTERNAL_MODULE__2__, __WEBPACK_EXTERNAL_MODULE__4__)
       ^
       |
       in this case we got window which does generate issue width server side rendering 
```

borracciaBlu:
Fix #1603

### DIFF
--- a/webpack.config.dist.js
+++ b/webpack.config.dist.js
@@ -4,11 +4,14 @@ var path = require("path");
 module.exports = {
   mode: "production",
 
+  target: "node",
+
   entry: "./src/index",
 
   output: {
     library: "Slider",
     libraryTarget: "umd",
+    globalObject: "this",
     path: path.join(__dirname, "dist")
   },
 


webpack.config.dist.js
```diff
@@ -4,11 +4,14 @@ var path = require("path");
 module.exports = {
   mode: "production",
 
+  target: "node",
+
```
akiran: I don't think we should set target as node

PR #1609, opened 2019-08-01
Check for innerSlider

aedenmurray:
Throws error if asNavFor.innerSlider does not exists. Happens when switching between pairs of maps (which I am looping over to build the slides) that do not have similar indexes.


tklkalok:
@aedenmurray My case also throws this error, It is fixed? Or I need to download this package and modify myself like what you changed?

PR #1612, opened 2019-08-04
Added option to use react-slick-styles to the readme

ivoilic:
I created the [react-slick-styles](https://github.com/ivoilic/react-slick-styles) package to supplement react-slick. I understand why these styles files weren't included in this package but it seems wasteful to install the whole slick package just for the styles. Having a cut down package with just the styles gives people that option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ npm install react-slick --save
 yarn add react-slick
 ```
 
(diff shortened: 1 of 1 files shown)


akiran:
Sorry we can't add this at the moment.

PR #1647, opened 2019-10-02
doc update: initialSlide is nowhere mentioned in API document

Dipen-Dedania:
It would be nice to add this in API doc as well as people won't know about it and they have to use ref and set with slickGoTo method to set it


akiran:
We will document initialSlide in the new docs

PR #1648, opened 2019-10-02
Fix: Added property validation check

Dipen-Dedania:
Validate the property for better debug mode


src/slider.js
```diff
@@ -208,3 +209,57 @@ export default class Slider extends React.Component {
     );
   }
 }
+
+Slider.prototypes = {
```
akiran: Typo here

It should be Slider.propTypes instead of Slider.prototypes

PR #1662, opened 2019-10-24
Fix issues with eslint, prettier, and tests

redbmk:
This fixes the eslint config so that it works properly with tests and lints more files besides just the `src` directory (now includes docs, examples, and tests).

This also makes it so prettier will run against jsx files on commit.

There were some failing tests before this with vague error messages, but something fixed those in the process.

This also commits `package-lock.json` to source control, since that's generally recommended. Let me know if it needs to be taken out gain for some reason, but I've personally run into issues before with minor updates to packages actually containing breaking changes, which is what `package-lock.json` helps prevent.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -15,16 +15,41 @@
     "node": true,
     "browser": true,
     "es6": true,
-    "jasmine": true
   },
   "parser": "babel-eslint",
   "plugins": [
       "react",
-      "import"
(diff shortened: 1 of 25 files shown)


redbmk:
@akiran I realize this is close to a year old now, but I'm curious why it was closed without any discussion.

PR #1711, opened 2020-01-15
Combine child.key and fallbackKey generating unique key

hullen:
When using row and lazyload, is generating an following error:
`Encountered two children with the same key, postcloned12.`

https://codesandbox.io/s/react-slick-slide-key-lazyload-esuk4


hullen:
I will close this PR, because i've made this on antd repository too and been merged.

PR #1723, opened 2020-02-04
fix: to handle image.onload event correctly

### LINKED ISSUE #1356: AdaptiveHeight does not re-calculate height on image load if multiple Sliders are on the page
I am having an issue where the adaptiveHeight setting is not re-calculating the height on a Slider when there are two Sliders on the page.

Here is codeSandBox that duplicates the issue but only sometimes and when the connection is slow enough: https://codesandbox.io/s/k27p314y6o . 

Locally the issue does not happen if I only have one Slider.  The Slide height is re-calculated properly once the image finishes loading.  However when a second Slider is added without images, and it is added higher on the page than  the slider with `adaptiveHeight={true}`, then the adaptiveHeight setting does not re-calculate the height of the Slider with the image (the Lower Slider on the page).   **This leaves much of the slide not visible as it extends outside the carousel**

NOTE:  I have looked into this some and have noticed that the `adaptHeight()` function is called when the image finishes loading but it is called on the wrong slider ( the first Slider on the page without the image).

choi2k:
Fix #1356 

`document.querySelectorAll(".slick-slide img")` selector will select all images under `.slick-slide` and add onload event on those images.  if there are multiple react-slick instance on the same page, the image.onload event may be handled by other react-slick instance . 

This commit aims to fix the issue by searching `".slick-slide img"` in the list reference instead of document.

### DIFF
--- a/src/inner-slider.js
+++ b/src/inner-slider.js
@@ -290,7 +290,7 @@ export class InnerSlider extends React.Component {
     });
   };
   checkImagesLoad = () => {
-    let images = document.querySelectorAll(".slick-slide img");
+    let images = this.list.querySelectorAll(".slick-slide img");
     let imagesCount = images.length,
       loadedCount = 0;
     Array.prototype.forEach.call(images, image => {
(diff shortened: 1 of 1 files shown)


beaukode:
You are right ! I got in same problem and i found same solution after some hours of debugging....
We need a merge...

Issue #1753, opened 2020-03-29
img tag, wrapped by div tag, breaks "centerMode" option with click event


jadenLog:
**The issue can be seen blow link:**
https://codesandbox.io/s/react-slick-playground-rqfxt

By Clicking the image of right or left, you can see the centering breaks.

The issue is occurred by image focus event which I found from "**react-slick.js**" - **line 786 ~ 797**.

```
        if (!image.onclick) {
          image.onclick = function () {
            return image.parentNode.focus();
          };
        } else {
          var prevClickHandler = image.onclick;

          image.onclick = function () {
            prevClickHandler();
            image.parentNode.focus();
          };
        }
```

Is there any reason that you added the focus event on the img tag?

PR #1756, opened 2020-04-01
#1753: Fix center break issue by focus event when slider Item has img…

jadenLog:
… tag.

Reference: https://github.com/akiran/react-slick/issues/1753

I fixed by blocking the focus event when "centerMode" with true.

I am guessing there is a reason that you focus img tag if there is img in the slider item.

If this is not a way you recommend, please let me know how you prefer to fix the issue.
I am willing to help. :)


jadenLog:
It has been 5 years... (I totally forgot that I made this PR on react-slick)

I was unable to reproduce the issue on the latest version. Therefore, I will proceed to close this PR.

Issue #1663, opened 2019-10-31
Unslick is not working


jorgefernandes:
I'm trying to set 'unslick' for desktop responsive, but slick is just erasing content inside slick wrapper

https://codesandbox.io/s/react-slick-playground-ylelk

cryptic022:
Hi, @jorgefernandes @akiran I have raised a PR for this issue. Please review.

akiran:
Thanks @cryptic022. Your PR fixed the issue.

Made a release react-slick@0.26.0 with the fix.

PR #1784, opened 2020-05-10
unslick default children return

cryptic022:
This PR will fix this [issue_1663](https://github.com/akiran/react-slick/issues/1663).
Due to parent settings variable, newChildren was always giving blank children. In the case of unslick, we should return direct children.


cryptic022:
Will raise again after fixing prettier issues

PR #1785, opened 2020-05-10
default children return in case of unslick

cryptic022:
This PR will fix this [issue-1663](https://github.com/akiran/react-slick/issues/1663)
Due to parent settings variable, newChildren was always giving blank children. In the case of unslick, we should return direct children.


akiran:
Thanks for PR @cryptic022

PR #1787, opened 2020-05-11
make sure "clickable" property is kept track off correctly

### LINKED ISSUE #1783: Wrapping slider in <a> tag doesn't work as expected on mobile
### CodeSandBox example

* CodeSandBox link: https://codesandbox.io/s/react-slick-playground-up7mf
* Example from the above link: https://up7mf.csb.app/

### Issue

If I wrap the entire `Slider` component in an `a` tag I can swipe/switch slides and click on any slide and go to the URL specified. This works fine on desktop. On mobile, however, it works, but only on the first slide. Once you swipe to another slide and tap, you are not taken to the link specified. On second tap, however, it does take you to the URL specified. 

I would expect to be taken to the URL specified on all slides on the first tap, not just one the first slide.

### Steps to reproduce the issue

Below are steps to reproduce the issue:
1. Go to https://up7mf.csb.app/ on a mobile device (or emulate a mobile device in your browser dev tools)
2. Tap on the first slide. Notice it takes you to Google.
3. Go back to https://up7mf.csb.app/
4. Swipe forward or backward (doesn't matter)
5. Tap on the new slide you are now on.
6. Notice you are not taken to Google this time.
7. Tap again.
8. Notice you are taken to Google now.

### Possible fix

This is the line of code that swallows the click event: https://github.com/akiran/react-slick/blob/ef32fce2bc9b4152d1006b661ed95512138be458/src/inner-slider.js#L413

This line marks the slide as not clickable:
https://github.com/akiran/react-slick/blob/ef32fce2bc9b4152d1006b661ed95512138be458/src/inner-slider.js#L453

I believe the `swipeMove` method that contains the line above should not get fired if the user has not actually moved but just performed a tap. Alternatively, if this method should be fired on tap, the slide should be marked as clickable again in a callback after `this.setState` on this line: https://github.com/akiran/react-slick/blob/ef32fce2bc9b4152d1006b661ed95512138be458/src/inner-slider.js#L455

### Conclusion

I can work on a PR if this gets acknowledged as a bug. Thanks.

bolatovumar:
fix #1783

### DIFF
--- a/src/inner-slider.js
+++ b/src/inner-slider.js
@@ -409,13 +409,14 @@ export class InnerSlider extends React.Component {
       this.slideHandler(targetSlide);
     }
   };
-  clickHandler = e => {
+  stopPropagationIfNotClickable = e => {
     if (this.clickable === false) {
       e.stopPropagation();
       e.preventDefault();
     }
     this.clickable = true;
   };
+  clickHandler = this.stopPropagationIfNotClickable;
   keyHandler = e => {
     let dir = keyHandler(e, this.props.accessibility, this.props.rtl);
     dir !== "" && this.changeSlide({ message: dir });
@@ -435,6 +436,11 @@ export class InnerSlider extends React.Component {
     window.ontouchmove = null;
   };
   swipeStart = e => {
+    /**
+     * Make sure we keep track if clickable property on each swipe.
+     * If we don't do this we end up with this bug: https://github.com/akiran/react-slick/issues/1783
+     */
+    this.stopPropagationIfNotClickable(e);
     if (this.props.verticalSwiping) {
       this.disableBodyScroll();
     }
--- a/src/utils/innerSliderUtils.js
+++ b/src/utils/innerSliderUtils.js
@@ -48,7 +48,7 @@ export const getSwipeDirection = (touchObject, verticalSwiping = false) => {
   xDist = touchObject.startX - touchObject.curX;
   yDist = touchObject.startY - touchObject.curY;
   r = Math.atan2(yDist, xDist);
-  swipeAngle = Math.round(r * 180 / Math.PI);
+  swipeAngle = Math.round((r * 180) / Math.PI);
   if (swipeAngle < 0) {
     swipeAngle = 360 - Math.abs(swipeAngle);
   }
@@ -195,7 +195,7 @@ export const slideHandler = spec => {
       finalSlide = animationSlide + slideCount;
       if (!infinite) finalSlide = 0;
       else if (slideCount % slidesToScroll !== 0)
-        finalSlide = slideCount - slideCount % slidesToScroll;
+        finalSlide = slideCount - (slideCount % slidesToScroll);
     } else if (!canGoNext(spec) && animationSlide > currentSlide) {
       animationSlide = finalSlide = currentSlide;
     } else if (centerMode && animationSlide >= slideCount) {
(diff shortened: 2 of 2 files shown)


src/utils/innerSliderUtils.js
```diff
@@ -48,7 +48,7 @@ export const getSwipeDirection = (touchObject, verticalSwiping = false) => {
   xDist = touchObject.startX - touchObject.curX;
   yDist = touchObject.startY - touchObject.curY;
   r = Math.atan2(yDist, xDist);
-  swipeAngle = Math.round(r * 180 / Math.PI);
+  swipeAngle = Math.round((r * 180) / Math.PI);
```
bolatovumar, line 51: I didn't add these brackets. It was the pre-commit hook.

src/utils/innerSliderUtils.js
```diff
@@ -195,7 +195,7 @@ export const slideHandler = spec => {
       finalSlide = animationSlide + slideCount;
       if (!infinite) finalSlide = 0;
       else if (slideCount % slidesToScroll !== 0)
-        finalSlide = slideCount - slideCount % slidesToScroll;
+        finalSlide = slideCount - (slideCount % slidesToScroll);
```
bolatovumar, line 198: I didn't add these brackets. It was the pre-commit hook.

src/utils/innerSliderUtils.js
```diff
@@ -265,7 +265,8 @@ export const changeSlide = (spec, options) => {
     slideOffset = indexOffset === 0 ? slidesToScroll : indexOffset;
     targetSlide = currentSlide + slideOffset;
     if (lazyLoad && !infinite) {
-      targetSlide = (currentSlide + slidesToScroll) % slideCount + indexOffset;
+      targetSlide =
```
bolatovumar, line 268: I didn't add these brackets. It was the pre-commit hook.

src/utils/innerSliderUtils.js
```diff
@@ -704,7 +715,7 @@ export const getTrackLeft = spec => {
       slideCount % slidesToScroll !== 0 &&
       slideIndex + slidesToScroll > slideCount
     ) {
-      slidesToOffset = slidesToShow - slideCount % slidesToScroll;
+      slidesToOffset = slidesToShow - (slideCount % slidesToScroll);
```
bolatovumar, line 718: I didn't add these brackets. It was the pre-commit hook.

src/inner-slider.js
```diff
@@ -409,13 +409,14 @@ export class InnerSlider extends React.Component {
...
   };
-  clickHandler = e => {
+  stopPropagationIfNotClickable = e => {
     if (this.clickable === false) {
       e.stopPropagation();
       e.preventDefault();
     }
     this.clickable = true;
   };
+  clickHandler = this.stopPropagationIfNotClickable;
```
akiran, line 419: Is this method name change only for readability ?

akiran: @bolatovumar

bolatovumar: @akiran this `stopPropagationIfNotClickable` is also used on line 443 below. If I just keep it named as `clickHandler` it would be weird calling a click handler directly from another method. This name change reflects more accurately what these lines of code do.

akiran: Hi, Adding this.clickable = true in swipeEnd method might be a better solution since we are setting clickable to false in swipeMove method.

Calling stopPropagationIfNotClickable in swipeStart is not correct.

bolatovumar: @akiran ok, i will try to add it to `swipeEnd`. I think I tried it already but it didn't work for me. I will try it agian.

bolatovumar: @akiran ok, i tried putting `this.clickable = true` into `swipeEnd` and it does fix this bug on mobile, however, it introduces another bug: now swiping on desktop becomes impossible. As soon as you swipe on desktop it takes you to the link specified. This shouldn't happen.

Having `this.stopPropagationIfNotClickable(e);` in `swipeStart` fixes the issue on mobile and doesn't introduce a new one on desktop.

PR #1793, opened 2020-05-12
Remove unused argument

### LINKED ISSUE #1792: Unused parameter passed to getOnDemandLazySlides function
On line 131 of `innerSliderUtils.js` there `spec` argument being passed-in that's unused. The `getOnDemandLazySlides` only accepts one parameter but two are being passed-in.

bolatovumar:
fix #1792

### DIFF
--- a/src/utils/innerSliderUtils.js
+++ b/src/utils/innerSliderUtils.js
@@ -48,7 +48,7 @@ export const getSwipeDirection = (touchObject, verticalSwiping = false) => {
   xDist = touchObject.startX - touchObject.curX;
   yDist = touchObject.startY - touchObject.curY;
   r = Math.atan2(yDist, xDist);
-  swipeAngle = Math.round(r * 180 / Math.PI);
+  swipeAngle = Math.round((r * 180) / Math.PI);
   if (swipeAngle < 0) {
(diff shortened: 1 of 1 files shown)


src/utils/innerSliderUtils.js
```diff
@@ -126,10 +126,7 @@ export const initializedState = spec => {
     currentSlide = slideCount - 1 - spec.initialSlide;
   }
   let lazyLoadedList = spec.lazyLoadedList || [];
-  let slidesToLoad = getOnDemandLazySlides(
-    { currentSlide, lazyLoadedList },
-    spec
-  );
+  let slidesToLoad = getOnDemandLazySlides({ currentSlide, lazyLoadedList });
```
bolatovumar, line 129: This is the real change I made. All other changes are automatic from the pre-commit hook.

akiran: @bolatovumar  This could be the correct fix
```let slidesToLoad = getOnDemandLazySlides({ ...spec, currentSlide, lazyLoadedList });```

Can you check and make a change in this PR

bolatovumar: @akiran done

PR #1804, opened 2020-05-20
fixed event collision between drag/swipe and click events

ChezFre:
When using `button` elements with an `onClick` the `onClick` event would always be fired after dragging due to the way event bubbling works. The `onClick` handler would be the first element to fire, before bubbling up to Slick Slider instead of the other way around.

This PR fixes that issue and also allows a user to define the minimum threshold a user should swipe/drag before invalidating click events on its items.

### DIFF
--- a/src/inner-slider.js
+++ b/src/inner-slider.js
@@ -687,11 +687,12 @@ export class InnerSlider extends React.Component {
     }
 
     const listStyle = { ...verticalHeightStyle, ...centerPaddingStyle };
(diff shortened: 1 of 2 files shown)


ChezFre:
I just checked with release 0.28.1, but it seems the solution implemented in https://github.com/akiran/react-slick/commit/f958da53d1f25e29bb43b5b7ad222621342092d8 does not change this behaviour. The `click` event is still executed when the mouse is released on end drag.

Issue #1809, opened 2020-05-27
current slide incorrect value when infinite=false and slidesToShow > 1


mgvez:
When multiple slides are shown and that the slider cannot move left anymore, the current slide index does not change. 

It seems that the state change that triggers animation uses the slide index to make the animation, and so does not change the value when animation is not possible. This is incorrect behaviour, as the clicked slide should be the current one, and not the leftmost visible slide. 

This is a problem when using a thumbnail slider asNavFor a larger slider, as it makes it impossible to highlight the currently selected slide.

This affect the slick-current css class, as well as the beforeChange and afterChange events.

Example here [CodeSandBox](https://codesandbox.io/embed/react-slick-playground-11rg7?fontsize=14&hidenavigation=1&theme=dark). To replicate, navigate to the last slides and click on any of the last ones. The red border is on .slick-current.

This probably also causes [1541](https://github.com/akiran/react-slick/issues/1541)

dkadrios:
I've` been using this as a temporary hack (in react)

```
onInit={() => {
  setTimeout(() => { ref.current.slickGoTo(0) }, 750)
}}
ref={(slider) => { ref.current = slider }}
```

Way far from ideal but works in the meantime

dkadrios:
Possible fix for this is in https://github.com/akiran/react-slick/pull/1811

nisargrthakkar:
any update on this??

aysegulkavaklii:
Same issue here +1

PR #1811, opened 2020-05-29
initialSlide not always honored; possible fix for issue #1809 

dkadrios:
When first loading, with initialSlide set to zero and slidesToShow greater than one, the internal currentSlide would become negative.  It's crucial that the intended first slide always be the actual first one visible.

It's possible that is only happens when in infinite mode.


nisargrthakkar:
+1

nut707:
Have some issues. Ant this workaround works fine. Should be merged!

PR #1823, opened 2020-06-10
Expose trackStyle prop in inner slider.

gjvpaet:
Expose the trackStyle prop to add inline styles to slick-track div.


jeromemiranda:
This is definitely needed to customize it more

akiran:
You can add css with classname slick-track

jeromemiranda:
I think this is needed other than the classname.

Considering you want to add multiple instances dynamically. Rather than adding class name one by one. Each easier to just pass the styles itself specially if you are using external css.

PR #1825, opened 2020-06-16
add hover callback prop from complex scene

lihanruiwork:
when hovered,I need this callback to do more things,thank you


akiran:
This needn't be in this library.

You can wrap a div on Slider and add a hover event to it.

PR #1833, opened 2020-07-07
Fix Auto Play - Clicked Dots Remains active #1432

### LINKED ISSUE #1432: Auto Play - Clicked Dots Remains active
In the Auto Play demo page (https://react-slick.neostack.com/docs/example/auto-play-methods).
When we click on a dot, the user is getting a feeling like two dots are active. 
Why are we having this behaviour ?

image

Caps-Looking:
Fix #1432 
As react-slick uses slick-carousel's css, changing the css doc's file wouldn't fix the error for library users, it would only fix for the docs.
I created a css file and a webpack loader to use it. This fix every use of the component.


webpack.config.js
```diff
@@ -23,6 +23,10 @@ module.exports = {
       {
         test: /\.md$/,
         loader: "html!markdown"
+      },
+      {
+        test: /\.css$/,
+        use: "css-loader"
```
hjJunior, line 29: I guess you have missed adding the package.json

src/css/styles.css
```diff
@@ -0,0 +1,10 @@
+.slick-dots li.slick-active button:focus:before,
+.slick-dots li button:hover:before
+{
+    opacity: 1;
+}
+
+.slick-dots li button:focus:before
+{
+    opacity: 0.25;
+}
```
hjJunior, line 10: Pls add a new empty line on the end of files

webpack.config.js
Caps-Looking: Nope, the css-loader is already on the package.json

akiran:
We can't use css file fix for this issue.

We will find a different solution for this problem

Issue #1180, opened 2018-04-09
Styles overwritten on application-owned div elements


nihlton:
demo: https://codesandbox.io/s/7yvq357xz6

for the given markup:  
```
<Slider {...settings}>
  <div style={slideStyle}>
    <img style={imageStyle} src="http://placekitten.com/g/400/200" />
  </div>
  ...
</Slider>
```
{slideStyle} will be ignored / overwritten.

mihauko:
Are those inline styles connected to additional div wrapper like on screenshot? What is the purpose of this div?

![selection_0961](https://user-images.githubusercontent.com/15020674/38719746-d6175f3c-3ef3-11e8-98ad-abe426c2e1ef.png)

It can be also visible on @nihlton demo.

I have this structure on 0.23.1 version

ViacheslavBurlaka:
@nihlton  have the same problem

nihlton:
@mihauko - was wondering that as well.

I suspect that they *intended* to apply those styles to the additional div wrapper, and are applying them to the application div instead.

nihlton:
any update on this issue?

dhruvramdev:
Facing Same Issue.

mopilo:
same here

charliedavison:
+1

tetogomez:
useful component but it's so hard to use with custom styles. :(

charliedavison:
Ended up wrapping an extra div around everything so we could prevent our styles from being overwritten. Would be nice if this was fixed properly though.

remingtonchan:
@charliedavison Hi, how was your approach in "wrapping everything"? I still can't get around this. 😅

pinglinh:
Any updates on this issue? 😭

goriverna:
As a workaround, you can use className on your div and define css class, in this way you can override css rules.
```
<Slider {...settings}>
  <div style={slideStyle} className='slick-custom'>
    <img style={imageStyle} src="http://placekitten.com/g/400/200" />
  </div>
  ...
</Slider>

```
and then create css rule on you css style:
```
.slick-custom{
    display: flex !important;
    justify-content: center;
}
```

akiran:
Fixed in react-slick@0.25.2

chrisneal:
@akiran We're still experiencing this issue in 0.25.2.

manuelgomezo:
same here

noobG:
yep still broken

rphly:
still broken

[...]

PR #1848, opened 2020-07-27
Removed additional wrapping <div> inside each slide

tazcarper:
Each slide has a `<div>` wrapper that is not needed. This looks to be originally used to wrap an array of elements. React offers the ability to use Fragments that replace the old `<div>` wrapper method. Replacing with a `<React.Fragment>` will eliminate the `<div>` wrapper and keep the same functionality. 

Fragments:
https://reactjs.org/docs/fragments.html 

Fixes issue related to this:
https://github.com/akiran/react-slick/issues/1180

Examples all working and all tests passing after Snapshot has been updated.


akiran:
@tazcarper This is a breaking change and might affect lot of existing users. So we can't accept it now.

We are planning to make a major version release where we bundle together all the breaking changes at once and release it as 2.0.

tazcarper:
I'd argue this is fixing a breaking change introduced in 23.0. The contents of a slide should be exactly what the direct children, or whatever is specified in the Slide setting, of the Slider contain. Currently, it takes the contents and wraps them with an additional `<div>`. The additional `<div>` causes numerous issues because it has no class and cant be directly manipulated / removed. Things like Flexbox, absolute position elements, ect become increasingly more difficult to work with. 

To my knowledge, this is unexpected behavior that isn't specifically mentioned in the documentation. nor is it reflected in the original Slick. If you don't want to remove the `<div>`, consider putting a className on the element so users can directly manipulate it in CSS. 

Thank you for your work.

akiran:
@tazcarper  Thanks for pointing this out.

I will consider this in upcoming releases.

estebano:
Maybe You @tazcarper could use a new setting/configuration option to change the current behavior. If new behavior would be configurable it will not be a breaking change anymore

danez:
Same here we are not able to update to any version >= 0.23 because of this. We have custom css classes that break when wrapped inside a div.

AgainPsychoX:
I have same issue as OP. The div is so pointless and annoying...

SoloJuanca:
It's odd this is still an issue

Issue #1838, opened 2020-07-16
focusOnSelect isn't animating to the correct slide


Iwi4a:
Bug: https://codesandbox.io/s/react-slick-playground-it0fw

**How to replicate:**
We have a carousel showing 3 slides with centerMode enabled. The first slide when you load the codesandbox is an image of a rabbit, if you select the slide on the left (an image of a dog, which is the last slide from the DOM), it slides to it as expected. From there, if you want to select the 2nd slide on the right (the image of the bear), instead of moving to the bear, it animates to the first slide (the rabbit).

**It is expected to animate to the correct slide**

Additional Details:
I tried investing it myself and I've got down to line 199 in innerSliderUtils, but not really sure how to fix it from there.
https://github.com/akiran/react-slick/blob/master/src/utils/innerSliderUtils.js#L199

It is not a slick issue, I have made a codepen for it:
https://codepen.io/iwicha/pen/MWKPJbZ

PR #1879, opened 2020-08-25
Fix for issue #1838

gautamz07:
This pull request fixes issue #1838.


gautamz07:
@akiran naughts ?

akiran:
I will review it this weekend

gautamz07:
@akiran , 

investigated this issue is bit , any obvious reason why you decided to use `targetSlide: previousTargetSlide` instead of just using `currentSlide` , which is already being reliably used in the changeSlide function ?

using `currentSlide` seems to get the functionality back on track. 

If you have a commit id for this change you introduced and the commit fixes 1 specific issue only , I can create a PR with a `git revert` commit and add the change. let me know.

gautamz07:
@akiran merging it anytime soon , let me know if I need to rebase.

PR #1885, opened 2020-09-01
Slide container type api bugfix

pupple-mint:
slide Api not working now, so I fixed it. plz merge


akiran:
This is not the correct intention of the slide prop.

PR #1898, opened 2020-10-04
removed buttons in non-infinite slider loop

Rahul1907:
So I used this slick slider in my recent project and I realised that when **I have kept prop infinite as false**, and when I am at First slide it still shows me the **Previous button (previos-arrow-button)** but actually that previous button will not do any thing. And same goes with the **Next Button (Next-arrow-button)** when it is on last slide it should not show the next button. Because it doesn't make sense to have that button as it wont be doing anything.

_In short what I mean to say is that there shouldn't be a button(Prev OR Next) when on clicking that it isn't going to show me another slide. **(this will be executed only if infinite prop is false)**_

I have made the required changes and also have attached the screenshot of responsive slide example.
Screenshot from 2020-10-04 13-16-41
Screenshot from 2020-10-04 13-17-42

### DIFF
--- a/src/inner-slider.js
+++ b/src/inner-slider.js
@@ -745,15 +745,28 @@ export class InnerSlider extends React.Component {
       listProps = { className: "slick-list" };
       innerSliderProps = { className };
     }
+    let showPrevBtn, showNextBtn;
+
+    if (this.props.infinite) {
+      showNextBtn = true;
+      showPrevBtn = true;
+    } else {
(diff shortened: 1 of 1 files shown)


akiran:
You can accomplish this with css.

Default behaviour of this library is to show disabled buttons on the last slide.
You can override that behaviour with css

Rahul1907:
ok thanks

PR #1904, opened 2020-10-09
Show the correct amount of items when navigating to previous page

gempir:
I have a Slider with 21 Items and show 6 slides per page. When navigating to a previous Page initially only 3 items are displayed. Because finalSlide becomes 18 and not 15 like it should be.

image

I can't find the reason this extra clause existed in the slideHandle function, I have tracked it back via git blame to this commit https://github.com/akiran/react-slick/commit/86b721896f1edbcac3c602fb6c28caf9047670c2 But there isn't much explanation why. 

With this change the correct amount of items are displayed.


gempir:
This isn't done. I have noticed a bug after 4 page switches, the animation glitches out and the items rapidly change after switching the page. Will come back later

Issue #1316, opened 2018-07-17
Sliding broken when slide content is wrapped in anchor tag


selrond:
[example on CodeSandbox](https://codesandbox.io/s/7y27y70no1)

When slide content is wrapped inside `<a />` tag, sliding behaves weirdly; unusable on FF and Safari, Chrome deals with it somehow. 

I believe this comes from here:

> `img` elements and `a` elements with an `href` attribute have their `draggable` attribute set to true by default.

[HTML spec - Drag and drop](https://html.spec.whatwg.org/multipage/dnd.html#drag-and-drop-processing-model)

Here's the code I think is related to this, with the proposed solution:

```js
export const swipeStart = (e, swipe, draggable) => {
  e.target.tagName === "IMG" && e.preventDefault();

  // my proposed solution
  e.target.tagName === "A" && e.preventDefault();

  if (!swipe || (!draggable && e.type.indexOf("mouse") !== -1)) return "";
  return {
    dragging: true,
    touchObject: {
      startX: e.touches ? e.touches[0].pageX : e.clientX,
      startY: e.touches ? e.touches[0].pageY : e.clientY,
      curX: e.touches ? e.touches[0].pageX : e.clientX,
      curY: e.touches ? e.touches[0].pageY : e.clientY
    }
  };
};
```

[source](https://github.com/akiran/react-slick/blob/master/src/utils/innerSliderUtils.js#L307)

What do you think?

franciscomorais:
you can solve this problem if you adds the draggable attribute setted by false.

selrond:
@franciscomorais try it on the codesandbox link I provided - it does not work

selrond:
this is not stale, the problem still occurs

Xb0X:
@selrond your fix is definitely needed !!! maybe you should open the PR :) If no I won't resist to do it 😄

selrond:
@Xb0X go ahead!

Mikhail-MM:
Is this fixed yet? Seeing this issue in production and not sure how to proceed. Is this only a React-Slick issue or Slick Carousel in general? I swear this type of formatting has worked when called .slick() in the jQuery implementation

ashraf-gemini:
@Mikhail-MM @selrond 

Here's a much cleaner fix, 

1. set the `draggable` attribute to `true` on the <a /> tag
2. and `onDragStart={(e) => e.preventDefault()}`


```jsx
<Slider {...sliderSettings}>
  <div>
[...]

PR #1907, opened 2020-10-14
Fix broken slide when slide content is wrapped in a tag. issue #1316

Xb0X:
Related to the issue from @selrond https://github.com/akiran/react-slick/issues/1316#issue-341951239 

As with IMG, we should be able to swipe when there is an A tag in the slide content. A tag became draggable when href is setted.


tranlehaiquan:
Why closed this? @Xb0X

Mikhail-MM:
@Xb0X @tranlehaiquan I am still seeing an issue with this behavior in production, this fix was never merged?

holgersindbaek:
Why wasn't this merged?

PR #1920, opened 2020-11-04
Feature/center infinite false

vladbdev:
I am currently working on a feature when I need the carousel to function like this:
- center mode, infinite false, no empty placeholder
- when user interacts with the carousel - click on next - set infinite true

In the current implementation there is a bug: if you set the initialSlide in the settings, it will break the arrow functionality for the first clicks on "next" - internally the slider starts targetSlide at 0, although the targetSlide should be populated from initialSlide - if provided.

Thank you,
Vlad


thanhlmm:
Tested with my production site. Everything is Ok

vladbdev:
@akiran can you please take a look and trigger a travis CI rebuild as the last one failed due to node-sass?

Best
Vlad

PR #1971, opened 2021-01-24
fix-1650

turelmert:
Removed preventDefault methods which causing "Unable to preventDefault inside passive event listener invocation." error from handlers of passive events: onTouchMove, onTouchStart. 

Related to https://github.com/facebook/react/pull/19654/files 

See "https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#parameters => passive"

### DIFF
--- a/src/utils/innerSliderUtils.js
+++ b/src/utils/innerSliderUtils.js
@@ -5,7 +5,7 @@ export function clamp(number, lowerBound, upperBound) {
 }
 
 export const safePreventDefault = event => {
-  const passiveEvents = ["onTouchStart", "onTouchMove", "onTouchEnd"];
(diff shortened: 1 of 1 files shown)


turelmert:
Noticed functional side effects after removing preventDefault methods in last commit. Implemented a new method to call preventDefault only if related event listener is not passive.

turelmert:
See https://github.com/facebook/react/blob/master/packages/react-dom/src/events/DOMPluginEventSystem.js between line 427-440

akiran (approved):

akiran:
Thanks for the PR.

Released 0.28.0 with this fix

monecchi:
Thanks guys! Upgrading to 0.28.0 fixed the issue!

PR #1986, opened 2021-02-09
Upgrade LazyLoad to functional component

Vogric:
The class component was updated by a functional component.


akiran:
Please create a single pull request converting all class component to functional components

Issue #1987, opened 2021-02-12
Two children with the same key warning when using row, lazyLoading and infinite.


lioryemini:
Hi,

I'm opening this issue as a sequel to a bug that has already opened #1422 by @houkanshan.
Unfortunately, this bug has been automatically marked as stale by a bot, because it has not had recent activity.

Reference for all details at #1422 

This issue also occurred to me (latest version - 0.28.0), and his solution solve it.
Can we add it to next version fix please? 
Can you please change the original bug status to "Open"?

Thanks!

sozdayka:
@akiran could you check PR, please

dm-kharkov:
Thanks @lioryemini and @sozdayka, it works for me!

@akiran Could You look on that PR please and merge that branch into the master ? 
Thanks!

VanTranTrucPhuong:
@akiran  please fix it from 2021 to now.

PR #2022, opened 2021-04-20
Fixed issue 1987

sozdayka:
Fixed issue: #1987 (Two children with the same key error when using row, lazyLoading and infinite.)

Author fix: @lioryemini


tuanbui050198:
please merge it, I have a problem with it. @akiran

PR #2029, opened 2021-05-07
Fix clicking next/prev arrows when initialSlide is not 0

### LINKED ISSUE #1946: Changing slide works incorrect when infinite is false and initialSlide > 1
Changing slide when initialSlide > 1 and infinite is false works wrong.
When I swipe right I expect to see next slide(initial + 1), but I always see second slide as next, when I swipe left, I always get first slide.
Also same behaviour when I click on arrows on this [example](https://codesandbox.io/s/ppwkk5l6xx?file=/index.js), but with initialSlide: 2, infinite: false.
I think that same behaviour we have always when initial slide is set and infinite setting is false.

[Link](https://stackblitz.com/edit/react-brc9ys?file=src%2FApp.js) to stackblitz to reproduce. https://stackblitz.com/edit/react-brc9ys?file=src%2FApp.js
Versions:
- react-slick: 0.27.13
- react: 17.0.1

jorrit:
This PR fixes #1946.

This bug got introduced with change 1c3d70cce795fe2daaeb4b57ce4013fd08b997aa. When `initialSlide` is set, `targetSlide` was not set to the right value. This is fixed.

Also, I think the intended change of 1c3d70cce795fe2daaeb4b57ce4013fd08b997aa can be implemented slightly simpler, by just checking if the new value is out of the allowed bounds.

### DIFF
--- a/__tests__/innerSliderUtils.test.js
+++ b/__tests__/innerSliderUtils.test.js
@@ -1,0 +1,33 @@
+import { changeSlide } from "../src/utils/innerSliderUtils";
+
+describe("changeSlide", () => {
+  it("should not pass to before 0 when not infinite", function() {
+    const settings = {
+      slidesToScroll: 1,
+      slidesToShow: 1,
+      slideCount: 10,
+      currentSlide: 0,
+      targetSlide: 0,
+      lazyLoad: false,
+      infinite: false
+    };
+
+    const targetSlide = changeSlide(settings, { message: "previous" });
+    expect(targetSlide).toEqual(0);
+  });
+
+  it("should not pass to after last slide when not infinite", function() {
+    const settings = {
+      slidesToScroll: 1,
+      slidesToShow: 1,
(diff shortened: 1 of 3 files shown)


apntzls:
Hello all, what's the status of this PR, I'm needing this fixed.

JoaoCSR:
Is there any way we can get a timeline on when to expect this merged?

yenienRey:
Please, merge this +1

hila-achieved (approved):

dramoturg:
Do you have any news?

akiran:
This issue is already fixed

gtb104:
If it’s already fixed, could you please point to the PR and release that addresses this issue? Ant design uses this library in their Carousel component. If there’s a specific version they should be at, then I’ll make a PR to pull in the correct version on their repo.

AttackXiaoJinJin:
come on, it's 2022 !!

PR #2061, opened 2021-07-16
Added VS Code Badge

anilloutombam:
To access code without cloning via VS Code


akiran:
We don't want to add it at the moment.

PR #2077, opened 2021-08-26
Replace classnames with clsx

wojtekmaj:
clsx is fully compatible with classnames API, is quickly gaining traction and is about 1/3rd smaller than classnames.

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
     "why-did-you-update": "^0.1.1"
   },
   "dependencies": {
-    "classnames": "^2.2.5",
+    "clsx": "^1.1.1",
     "enquire.js": "^2.1.6",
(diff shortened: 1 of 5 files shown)


akiran:
We will revisit this later.

dartess:
@akiran but why? It is Drop-in replacement. No testing or inspection required. Thus, you just need to release a patch. At the same time, the new library weighs less. For example, `clsx` uses `@mui/*`.

jorrit:
@akiran 
I hope you can reopen this PR and merge it. clsx is clearly the preferred package these days and is more likely to be already available in a project that uses react-slick.
Another solution would be to rewrite react-slick to not use any of these packages and build css classes by hand.

image

Issue #2075, opened 2021-08-25
RTL issue on first slide change when infinite is set to false


hacktui:
Hi,

slider does not work well on the first slide change with setting:
rtl: true,
infinite:false

slide start at 3(which is correct) but when pressing [previous] button the first time, the slide will go straight to slide 1 instead of 2.
https://codesandbox.io/s/react-slick-playground-forked-dpxze

Vandrosov99:
Hi @akiran can you check this pull request for this problem? 

https://github.com/akiran/react-slick/pull/2125

kkmitaBSG:
some solution ?

OsDroidi:
having the same issue, any solution pls? @akiran

Sanjay1398:
I'm also facing this issue any solution?

PR #2125, opened 2022-02-01
fix order view with rtl mode

Vandrosov99:
Hi, can you review/check my solution to this problem?
[https://github.com/akiran/react-slick/issues/2075](url)


grignards:
I'm interested in this feature too, could I help to review it?

akiran:
I am not sure about this fix.

We will fix the RTL issues in a separate PR soon

PR #2144, opened 2022-04-03
Add pnpm to readme

giovannipds:


giovannipds:
@akiran why closing it without feedback?

PR #2192, opened 2022-10-12
Feature : ondemandSlidesToLoad props added to control slides loading on Lazyload (ondemand)

RifatCse:
Feature, Now users can control how many slides they want to load on lazy-load ondemand functionality.
default value : 0.


RifatCse:
@akiran please accept my request! we need this change.

PR #2218, opened 2023-02-15
Color of dots can now be changed.

Jastagar:
Hello, I have enabled the ability to change the color of the dots using 'dotsColor' setting. You can use this setting to change the color as per CSS colors. I was not sure to how to add tests to this change hence as per your guild lines I was unable to include, but please be sure to ping me incase anything goes wrong at all. (also this is my first ever contribution to an open source project so, please try to ignore any unintentional inconvenience I might have caused )


Jastagar:
use dotColor prop to change the dot's colors

akiran:
We can't add this new setting as it is not part of the slick carousel API.

You can change the color of the dots with custom css.

Issue #2206, opened 2022-12-23
Element type is invalid: expected a string (for built-in components) or a class/function...


PremaYT:
I am using react-slick as a dependency in one of my custom react package. Slider works fine in the component whereas it throws below error when we publish the component as a package and use it in another component. 

Error : 
###  Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: object. Check the render method 'SimpleSlider'.

```
import Slider from "react-slick";

const SimpleSlider = () => {
    const settings = {
      dots: true,
      infinite: true,
      speed: 500,
      slidesToShow: 1,
      slidesToScroll: 1
    }
    return (
      <div>
        <Slider {...settings}>
          <div>
            <h3>1</h3>
          </div>
          <div>
            <h3>2</h3>
          </div>
          <div>
            <h3>3</h3>
          </div>
          <div>
            <h3>4</h3>
          </div>
          <div>
            <h3>5</h3>
          </div>
          <div>
            <h3>6</h3>
          </div>
        </Slider>
      </div>
    );
}

export default SimpleSlider;
```

fredericrous:
I can reproduce this exact issue.
the people there too => https://github.com/akiran/react-slick/issues/2118

jordisantamaria:
Could avoid this error by using the component like this:

import Slider from "react-slick";

 <Slider.default>
 </Slider.default>

bluesbroz:
Just had the same problem.
Using vite react-ts.
Running in dev mode and building was fine, but when i tried to run ssr i was getting that error.

The imported object was `{ default: Slider }` instead of just `Slider` even if i tried to import like this `import { default as Slider } from 'react-slick'`.

Fixed with this:
```
// @ts-ignore
const SliderComponent = typeof window === 'undefined' ? Slider.default : Slider;
```

Maybe this is just an issue of my implementation of ssr.

naghtigal:
> ```
> // @ts-ignore
> const SliderComponent = typeof window === 'undefined' ? Slider.default : Slider;
> ```

Thanks, it did a trick.

Alevale:
Still an issue, at least for us on the version `0.29.0`

The fix from @paolozanotti-f2 https://github.com/akiran/react-slick/issues/2206#issuecomment-1961173249 worked though, so thank you.

qnauman:
> Could avoid this error by using the component like this:
> 
> import Slider from "react-slick";
> 
> <Slider.default> </Slider.default>

[...]

PR #2258, opened 2023-07-15
fix: #2206 Element type is invalid

fredericrous:
allow react-slick to be bundled into a library


fredericrous:
Fix suggested for #2206 was not merged in 3 years. closing to cleanup my list of open PR

PR #2259, opened 2023-07-19
Update inner-slider.js

jbbpatel94:
updated hardcoded dir value based on rtl props


akiran:
We have a lot of RTL issues that need to be fixed. We will handle this along with those fixes

PR #2262, opened 2023-07-31
RTL-fix-1

pratyushbh:
fixed some issues regarding RTL


Noushad-web:
@pratyushbh Thanks for this PR; it helped me a lot! However, with the `infinite: true` slide setting, it starts from slide -2 instead of 0.
I’ve included a picture below to illustrate this.

Screenshot 2025-01-04 at 5 20 09 PM

PR #2281, opened 2023-09-11
Update slick-theme.css

Ayush472:
I am having an issue where I am trying to disable the prev and next button by applying the css. But the css we are applying is not working and after having a glance at your plugin code we found that it is applied to ::before and ::after pseudo elements. we want you to update the css to apply on disbled class (without pseudo) as demonstrated below:

remove below:
.slick-prev.slick-disabled:before,
.slick-next.slick-disabled:before
{
    opacity: .25;
}

add the following:
.slick-disabled{
  filter: grayscale(1);
}

### DIFF
--- a/docs/slick-theme.css
+++ b/docs/slick-theme.css
@@ -94,7 +94,9 @@
 {
     content: '→';
 }
-
+.slick-disabled{
+    filter: grayscale(1);
+  }
 .slick-next
 {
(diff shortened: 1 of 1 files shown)


akiran:
These css files are only for our docs.

You should use slick-carousel package for CSS in your projects

PR #2288, opened 2023-10-01
fix: set slider wrapper key based on children keys passed

jamessspanggg:
### Context

Currently `react-slick` does not support dynamic children, where if any of the children is updated in run time, the `react-slick` slider does not re-render accordingly. This is due to the wrapper component having it's own `key` calculation that does not take into account the children's keys that was passed.

### Solution

Utilise a combination of the children's `key` for the wrapper, this ensures that if any of the children's key updates, the wrapper knows to re-render accordingly.

### DIFF
--- a/src/slider.js
+++ b/src/slider.js
@@ -171,9 +171,10 @@ export default class Slider extends React.Component {
             currentWidth = children[k].props.style.width;
           }
           if (k >= children.length) break;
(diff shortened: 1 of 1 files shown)


jamessspanggg:
@akiran appreciate it if you could take some time out to look at this fix for the PR as it might help out other people that uses ur package 🙏

Issue #2300, opened 2023-10-31
Add an option to position dots above or below the main slider


T-Fletcher:
# Situation

I'm replicating a slideshow widget using React Slick to serve slides containing images wrapped in links.

This functionality could be hand-rolled but given my project already use React Slick elsewhere and the control it provides, I've gone down this path. 

The UI I need to copy has the dots/slide group titles above the slides themselves i.e. pseudocoded:

```
group 1 | group 2

group 1
 - link 1
 - link2

group 2
 - link 1
 - link2
...
```

In my component, I've used `appendDots` and `customPaging` to insert text from slide group headings into the Dots as navigation:

```javascript
const mainSliderSettings = {
    accessibility: true,
    adaptiveHeight: true,
    arrows: false,
    autoplay: false,
    dots: true,
    fade: true,
    className: 'slick-tiles',
    infinite: false,
    slidesToScroll: 1,
    slidesToShow: 1,
    appendDots: (dots) => buildNavWrapper(dots),
    customPaging: (i) =>
      node.relationships?.field_tile_image_groups[i].field_title ? (
        <div
          className="section__nav-item"
          key={`${sectionTitle && convertStringToId(sectionTitle)}-pager-${i}`}
        >
          <a data-slide-index={i} href="#">{node.relationships?.field_tile_image_groups[i].field_title}</a>
        </div>
      ) : (
        <></>
      ),
  };

...

const buildNavWrapper = (dots) => {
    return (
      tileGroupLength() > 1 && (
        <nav className="section__nav-wrapper t-center">
          <ul
            className="section__nav text-dots"
            id={
              `${sectionTitle && convertStringToId(sectionTitle)}-pager` ||
              `slider-${Math.ceil(Math.random() * 10000000)}`
            }
            style={{
              position: 'relative',
              left: '-50%',
            }}
          >
            {dots}
          </ul>
        </nav>
      )
    );
  };
```

Which results in:

```html
<div class="slick-slider slick-tiles slick-initialized" dir="ltr">
  <div class="slick-list" style="height: 715px">
    <div class="slick-track">...slides here</div>
  </div>
  <nav class="slick-dots">
    <ul
      class="section__nav text-dots"
      id="tile-images-pt-heading-show-all-enabled-pager"
    >
      <li class="slick-active">
        <a data-slide-index="0" href="#">Tile images group 1</a>
      </li>
      <li class="">
        <a data-slide-index="1" href="#">Tile images group 2</a>
      </li>
[...]

PR #2301, opened 2023-10-31
Added dotsAboveSlider prop

T-Fletcher:
Addresses [issue 2300](https://github.com/akiran/react-slick/issues/2300)

Adds new prop to determine rendering order of the dots and slider. Defaults to `false`


akiran:
We can't add new props.

Commented the issue with solutions for this usecase

PR #2313, opened 2023-12-18
Create asNavForFunctionComponent.js

momadejunior:
I intend to add a asNavFor, instead of class component use function component, I believe having two version of this component would be nice


Arunav1:
Hey @momadejunior can I be assigned with this issue as I would like to solve this issue..

akiran:
Examples are already converted to function components

PR #2318, opened 2023-12-24
feat: Add dynamic slide adjustment to multi-item carousel

sandeepmallina:
**Enhancement Overview:** I have added two buttons, “Add Slides” and “Remove Slides”, to the multi-item carousel container. These buttons allow users to dynamically adjust the number of slides shown in the carousel.

**Justification:** The motivation behind this enhancement is to provide users with a more interactive experience and a clearer understanding of how the carousel behaves with different numbers of slides. By manipulating the number of slides, users can see how the carousel adjusts, providing a more tangible understanding of the slidesToShow and slidesToScroll properties.

### DIFF
--- a/examples/MultipleItems.js
+++ b/examples/MultipleItems.js
@@ -1,17 +1,38 @@
-import React from "react";
+import React, { useState } from "react";
 import Slider from "../src/slider";
 
 function MultipleItems() {
+  const [countOfSlides, setCountOfSlides] = useState(3);
   const settings = {
     dots: true,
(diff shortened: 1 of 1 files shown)


akiran:
MultipleItems is a basic example to illustrate that multiple slides can be shown with carousel. 
Don't want to complicate it with add and remove buttons.

We have dynamic slides example to illustrate your usecase
https://react-slick.neostack.com/docs/example/dynamic-slides

PR #2326, opened 2024-01-19
fix:  height style issue

MinhOmega:
* fix: style issue when the number of slides is less than or equal to slidesToShow

* fix: only modify height & fix wrong cloned element

* feat: add tests

* https://github.com/ant-design/react-slick/pull/107

* https://github.com/ant-design/ant-design/issues/44766


akiran:
This was fixed in 0.31.0 release

Issue #2315, opened 2023-12-20
slickPrev not working after refresh starting v0.27.9


digitalartcrew:
## Issue

Starting version v0.27.9 `slickPrev` stop working after refresh. `targetSlide` values becomes `-1` instead of previous slide index. See below:

The screenshots demonstrate a change slide action starting from index 3 in both versions.

Not working as of 0.27.9
![0 27 9-slickPrev-after-refresh](https://github.com/akiran/react-slick/assets/10134023/ca7f3120-dae3-4c25-a054-55ce6ce1ab2a)

Working 0.27.8 and below
![0 27 8-slickPrev-after-refresh](https://github.com/akiran/react-slick/assets/10134023/421f306a-77ae-49a9-a09c-6873b63c8b12)

It looks like there is a bug here, regardless if lazyload is true/false it will also be overwritten when infinite is false
![Screenshot 2023-12-20 at 1 13 54 PM](https://github.com/akiran/react-slick/assets/10134023/563b8b3c-f45a-4f5f-8bc8-cd5f91b9b34d)

This code was also added when version failed `git diff 69348c1 a43f109`
![Screenshot 2023-12-20 at 4 14 37 PM](https://github.com/akiran/react-slick/assets/10134023/07414289-f118-418c-a883-787cd78a7517)

digitalartcrew:
@akiran Would you mind taking a look at this issue? Thanks

akiran:
This could be because of unnecessary re-renders of Slider when state in a parent component changes.

I am working on a fix and I will make a release soon
https://github.com/akiran/react-slick/pull/2331

akiran:
@digitalartcrew Can you give a code sample to replicate the issue you are facing?

digitalartcrew:
> This could be because of unnecessary re-renders of Slider when state in a parent component changes.
> 
> I am working on a fix and I will make a release soon #2331

Thank you very much. It looks like this issue is resolved!

PR #2327, opened 2024-01-23
2315 slick prev refresh

digitalartcrew:
[#2315 ](https://github.com/akiran/react-slick/issues/2315)

After refresh the value of `previousTargetSlide` is currently resetting to 0. I replaced with the value of `currentSlide` which state value persist as same value 0f `previousTargetSlide` before refresh.

Before refresh:

Screenshot 2024-01-22 at 10 43 47 PM

After refresh:

Screenshot 2024-01-22 at 11 10 00 PM


akiran:
I don't think this is the fix for the issue you mentioned.

Issue #2328, opened 2024-01-24
May I develop it so that infinite swipe is possible when the values of 'slidesToShow' and 'showElement' are the same, and 'infinite' is set to true?


rjsdnql123:
Hello. I'm using the `react-slick` library successfully.

The requirements for my project are to have 9 `items(element),` with `slidesToShow` set to 9 so that all elements are visible on the screen, and `infinite` is set to `true.`

The expected outcome is to have all elements visible on the screen and be able to slide left and right.

However, I've confirmed that it's not currently functioning as expected.

Looking at the issue list, it seems there are others with similar requirements. Would it be okay for me to try making modifications in this regard?

## Related issues

https://github.com/akiran/react-slick/issues/2102

## before
<img width="1258" alt="image" src="https://github.com/akiran/react-slick/assets/61114705/3d5ad882-a508-4794-aa79-003fe62bf8d0">

## after 

<img width="687" alt="image" src="https://github.com/akiran/react-slick/assets/61114705/3382a6e1-f6f3-4c0f-8e5b-641e12715b9a">

### Guidelines for posting a new issue

* Please replicate your issue with this [CodeSandBox](https://codesandbox.io/s/ppwkk5l6xx) and provide a link to it along with the issue description

rjsdnql123:
+ add https://github.com/akiran/react-slick/issues/1813

rjsdnql123:
@akiran 
If you could check, I will summarize and post my changes. Thank you for creating a great library! :)

akiran:
Sure. You can send a PR

rjsdnql123:
https://github.com/akiran/react-slick/pull/2329%20%20%20%20Could%20you%20please%20review%20my%20pull%20request?%20:%20)

akiran:
Merged #2329

rjsdnql123:
Thank you 😄

PR #2329, opened 2024-01-24
Fix: If the element (DOM element) and the value of 'slidesToShow' are…

rjsdnql123:
If the element (DOM element) and the value of 'slidesToShow' are… the same, and 'infinite' is set to true, swipe should be possible.

https://github.com/akiran/react-slick/issues/2328

### DIFF
--- a/src/slider.js
+++ b/src/slider.js
@@ -198,7 +198,7 @@ export default class Slider extends React.Component {
     if (settings === "unslick") {
       const className = "regular slider " + (this.props.className || "");
       return <div className={className}>{children}</div>;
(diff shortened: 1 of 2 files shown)


akiran (approved):

rjsdnql123:
@afc163 
@rodrigo-arias
I'm sorry for the delay in checking.

I realized the bug late because it differs from the way I usually use it.

In this pull request, the bug occurred due to the omission of the unslick property in filterSettings, which initializes and passes properties.

https://github.com/akiran/react-slick/issues/2315


Resolved
https://github.com/akiran/react-slick/pull/2366

